PR #10, opened 2014-08-01
Remove test_bucket_recreate_not_overriding

gaul:
This test fails on AWS in regions other than us-standard.  See #9 for
more discussion.


samuelsh:
@alfredodeza , @yehudasa This test will always fail according to AWS docs:
http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html
> Your previous request to create the named bucket succeeded and you already own it. You get this error in all AWS regions except US East (N. Virginia) region, us-east-1. In us-east-1 region, you will get 200 OK, but it is no-op (if bucket exists it Amazon S3 will not do anything). 

I guess this pull requests need to be approved

gaul:
@alimaredia @cbodley please review.

PR #20, opened 2014-08-30
test_object_copy_to_itself should succeed

gaul:
AWS-S3 allows copying an object to itself, even without a change in
metadata.

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -6129,10 +6129,11 @@ def test_object_copy_to_itself():
 
     copy_source = {'Bucket': bucket_name, 'Key': 'foo123bar'}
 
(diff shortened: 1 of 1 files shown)


andrewgaul:
Actually, Amazon [specifies](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html) rejecting self-copy without replace:

> You cannot copy an object to itself unless the MetadataDirective header is specified and its value set to REPLACE.

As above, testing reveals a different behavior.  Should we align the test with the specification or the implementation?

yehudasa:
@andrewgaul @alfredodeza is rgw accepts, and aws accepts? We can test both behaviors, mark 'fails_on_rgw', and 'fails_on_aws' where appropriate. Not sure I like this solution.

PR #44, opened 2015-03-10
Add fails_on_mod_proxy_fcgi to functional.test_s3.test_100_continue

andrewschoen:
mod_proxy_fcgi only has partial 100 continue support so we want to skip this test when testing against it.

Signed-off-by: Andrew Schoen <aschoen@redhat.com>

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -4673,6 +4673,7 @@ def _simple_http_req_100_cont(host, port, is_secure, method, resource):
 @attr(operation='w/expect continue')
 @attr(assertion='succeeds if object is public-read-write')
 @attr('100_continue')
+@attr('fails_on_mod_proxy_fcgi')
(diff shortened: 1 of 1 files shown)


yehudasa:
@andrewschoen we should also have this backported to hammer and firefly. Can you add 'Backport: hammer, firefly' tag?

andrewschoen:
@yehudasa do you mean just add a label to this PR saying 'Backport; hammer, firefly'?

I'm using the firefly branch for testing, it also needs 992f2f169b1b61579dc7a3973b819e8c6bdb3587

yehudasa:
@andrewschoen I mean add it to the commit message itself. Hmm, didn't remember that other commit, will need to have that one on all branches too.

andrewschoen:
@yehudasa I've modified the commit message, let me know if that's not what you meant.  Thanks.

PR #54, opened 2015-05-15
bootstrap: yum install the list of dependencies

ktdreyer:
This optimizes the bootstrap section for RPM in a couple ways:

1. There's no need to check "`rpm -q`" before doing "`yum install`". Yum will gracefully skip packages that are already installed.
2. Just assume that the user has sudo rights; don't "echo" what they should do. If there are problems, they can read the bootstrap shell script to figure out what's going on.
3. Handle all the packages at once; don't loop through them individually.

Related: http://tracker.ceph.com/issues/11637

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -15,16 +15,7 @@ if [ -f /etc/debian_version ]; then
     fi
 fi
 if [ -f /etc/redhat-release ]; then
-    for package in python-pip python-virtualenv python-devel libevent-devel; do
-        if [ "$(rpm -qa $package 2>/dev/null)" == "" ]; then
-            missing="${missing:+$missing }$package"
-        fi
-    done
-    if [ -n "$missing" ]; then
-        echo "$0: missing required packages, please install them:" 1>&2
-        echo "  sudo yum install $missing"
-        exit 1
-    fi
+    sudo yum install python-pip python-virtualenv python-devel libevent-devel
 fi
 
 virtualenv --no-site-packages --distribute virtualenv


andrewschoen:
Why not do the apt work here as well?  It's almost the exact same change.

bootstrap
```diff
@@ -15,16 +15,7 @@ if [ -f /etc/debian_version ]; then
...
-        if [ "$(rpm -qa $package 2>/dev/null)" == "" ]; then
-            missing="${missing:+$missing }$package"
-        fi
-    done
-    if [ -n "$missing" ]; then
-        echo "$0: missing required packages, please install them:" 1>&2
-        echo "  sudo yum install $missing"
-        exit 1
-    fi
+    sudo yum install python-pip python-virtualenv python-devel libevent-devel
```
andrewschoen: Do we need a -y here?

andrewgaul:
:-1: How will this script work when the user lacks sudo privileges, e.g., Travis CI container-based builds:

https://github.com/andrewgaul/s3proxy/blob/master/.travis.yml
https://github.com/andrewgaul/s3proxy/blob/master/src/test/resources/run-s3-tests.sh

andrewschoen:
@andrewgaul Could we have the install fail silently and print the old message saying the packages need installed if so?

andrewgaul:
@andrewschoen As long as `bootstrap` continues to support non-root users this satisfies my use case.  However, what problem does this commit solve (bug has private access)?  Instead should you have `install-apt-dependencies.sh` and `install-yum-dependencies.sh` scripts which your external test harness invokes?

andrewschoen:
@andrewgaul Our configuration management had typically installed these packages on our testnodes.  However, we're trying to push the responsibility of installing things back out to the projects or tasks that need them.

Our current test harness only invokes ``bootstrap``, but I suppose we could change that teuthology task to install these packages before calling ``bootstrap`` in this project.

bootstrap
ktdreyer: Shoot, yes. Will amend and re-push.

ktdreyer:
@andrewgaul , thanks for pointing out the use case for users who don't have the packages installed and don't have sudo privs. I wasn't aware that s3-tests was used in that fashion, but it certainly makes sense.

The main problem I wanted to solve was to make bootstrap automatically install the dependencies if they are not available, rather than simply bailing with an error text. I'll amend this PR to only include this change. I'm not as concerned about the other optimizations at this point.

(Also, apologies for the private ticket access. As @andrewschoen made reference to above, it's a request from @vasukulkarni to install `python-devel` and `python-virtualenv` throughout our downstream lab in order to satisfy this test suite. @andrewschoen and I were trying to see if we could avoid installing those packages globally just for this one test, hence this `bootstrap` change.)

ktdreyer:
I'm going to just close this PR for the newer one at https://github.com/ceph/s3-tests/pull/55

PR #65, opened 2015-07-08
Add a test for concurrent PUT requests.

timuralp:
Adds two tests for concurrent PUT requests and concurrent overwrite
PUT requests. The expectation is that in each case the request with
the latest submitted time should be the one that is externally
visible.


andrewgaul:
This new test fails for me sometimes when running against aws:

```
FAIL: s3tests.functional.test_s3.test_concurrent_object_overwrite
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gaul/work/s3-tests/virtualenv/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/gaul/work/s3-tests/s3tests/functional/test_s3.py", line 988, in test_concurrent_object_overwrite
    eq(measured_size, small_size + 40)
AssertionError: None != 1048616
```

andrewgaul:
Reference for last writer wins: https://forums.aws.amazon.com/message.jspa?messageID=72266

timuralp:
I increased the sleep time between kicking off the large upload thread and when the small upload occurs to 1s. Ideally, we would like to pause the upload process and inject the second request. However, I could not figure out how to do that. Attempt to use the ```FakeFileWriter``` showed that it's not quite meant for that, as I observed the request created in the interrupt on the wire before the request that triggered the interrupt.

yehudasa:
@timuralp @andrewgaul it looks like a racy test, I'm not sure it's the correct way to test that. Moreover, don't we already have a test that mimics that in a different way?

timuralp:
@yehudasa I updated the test to use the ```FakeWriteFile``` primitive to induce the desired behavior. The content MD5 value needs to be supplied ahead of time, as otherwise the interrupt would be called prematurely.

I also couldn't find tests that test exactly the same behavior (but I haven't read all of them). I assumed that any test that does this would have to either pre-compute MD5 or use threads (like the prior iteration of the PR) and no tests appear to test concurrent PUTs in that way. Skimming through the list of all tests in ```functional/test_s3.py``` I also couldn't find a test that does this. Could you point me at a test that validates this behavior?

yehudasa:
@timuralp the s3tests-test-readwrite in this repository are doing concurrent writes, and ir doesn't need to have the md5 provided ahead of time (md5 is embedded in the data).

timuralp:
@yehudasa updated the PR to remove MD5 generation and the import statement for hashlib.

PR #66, opened 2015-07-14
Remove "--allow-hosts None" because it magically makes things work

dmick:
Fixes: 12302
Signed-off-by: Dan Mick <dan.mick@redhat.com>


yuriw:
@dmick  We need it in hammer and possibly in other branches, a?

PR #70, opened 2015-08-03
Fix the example which is deprecated in README.rst

tobegit3hub:
That case doesn't exist any more. And I change it to a workable one.

Signed-off-by: tobe <tobeg3oogle@gmail.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,7 +73,7 @@ Once you have that, you can run the tests with::
 
 You can specify what test(s) to run::
 
-	S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests.functional.test_s3:test_bucket_acl_grant_userid_read
+	S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests.functional.test_s3:test_bucket_list_empty
 
 Some tests have attributes set based on their current reliability and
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -73,7 +73,7 @@ Once you have that, you can run the tests with::
 
 You can specify what test(s) to run::
 
-	S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests.functional.test_s3:test_object_acl_grant_public_read
+	S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests.functional.test_s3:test_bucket_acl_grant_userid_read
```
andrewgaul: :+1: although could you pick a simpler test case like `test_bucket_list_empty` which is more likely to run correctly?  This test requires less user configuration and some S3 clones lack support for ACLs.

tobegit3hub: But I'm not sure if it's what we want. The case here tells how to specify the test instead of all the tests. I will update the pr if everyone wants that :smiley:

andrewgaul:
`s3tests.functional.test_s3:test_bucket_list_empty` is a single test and the most simple of all of them.

andrewgaul:
:+1:

tobegit3hub:
Thanks @yehudasa for merging this :smiley:

PR #75, opened 2015-09-02
Test multipart copy part feature

jmunhoz:
Add multipart uploads copy part feature tests.

See https://github.com/ceph/ceph/pull/5673

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -4555,6 +4555,14 @@ def transfer_part(bucket, mp_id, mp_keyname, i, part):
     part_out = StringIO(part)
     mp.upload_part_from_file(part_out, i+1)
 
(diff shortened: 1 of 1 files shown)


andrewgaul:
:+1:

jmunhoz:
@yehudasa I will take care of the missing test case

jmunhoz:
@yehudasa missing test case added at https://github.com/ceph/s3-tests/pull/86

jmunhoz:
@robbat2 thanks for reporting, see https://github.com/ceph/ceph/pull/5673 to details. The 'multipart copy part' feature did not ship with Jewel at the end. Those tests are part of the effort to support that feature. With the feature in place the -ERR_NOT_IMPLEMENTED code disappears and tests pass.

As suggested, I think it could make sense adding the fails_on_rgw annotation (PR https://github.com/ceph/s3-tests/pull/99) on test_multipart_copy_small and test_multipart_copy_multiple_sizes while the feature is not ready. What do you think @yehudasa?

PR #76, opened 2015-09-10
Test RequestId on error

jmunhoz:
Add RequestId on error tests

See https://github.com/ceph/ceph/pull/5866


jmunhoz:
@yehudasa would it be possible merging those two tests? https://github.com/ceph/ceph/pull/5866 was merged several weeks ago.

jmunhoz:
@yehudasa you're right, e.request_id comes from x-amz-request-id instead of the exception body. I updated the two patches to use values coming from the header and the body. I think they are ok now

PR #79, opened 2015-10-07
Improve proxy error support: "Bad Request" case

robbat2:
If you are run the S3 testsuite against a reverse proxy (eg haproxy)
rather than RGW directly, some of the tests will return errors from the
proxy, rather than haproxy itself.

Some proxies differ from RGW in the exact case of 'Bad Request', so do
the match in a case-insensitive manner. Haproxy for example returns 'Bad
request'.

This removes the need for some of the prior fails_on_dho test tags, as
the failure was due to haproxy.


yehudasa:
@robbat2 branch conflicts

robbat2:
Updated for master

PR #84, opened 2015-10-15
rgw: add AWS4 auth testing support

jmunhoz:
This patchset adds AWS4 auth testing support.

### DIFF
--- a/s3tests/functional/test_headers.py
+++ b/s3tests/functional/test_headers.py
@@ -11,11 +11,13 @@ import random
 import string
 import socket
 import ssl
+import os
 
 from boto.s3.connection import S3Connection
 
 from nose.tools import eq_ as eq
 from nose.plugins.attrib import attr
+from nose.plugins.skip import SkipTest
 
 from .utils import assert_raises
 import AnonymousAuth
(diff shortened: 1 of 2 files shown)


yehudasa:
@jmunhoz my main issue here is whether there's a way to have all common tests run twice, for both aws2 and aws4.

jmunhoz:
@yehudasa I changed the approach to keep all headers test cases (common, aws2 and aws4) in the original file so the current behavior is running common and aws2 test cases but skipping the specific aws4 test cases as expected.

andrewgaul:
@jmunhoz  How do I run AWS4 auth tests?  When setting `host = 127.0.0.1` and `region = use-east-1` I encounter the same symptoms as boto/boto#2624:

```
$ S3TEST_CONF=s3proxy.conf S3_USE_SIGV4=true ./virtualenv/bin/nosetests
...
UnboundLocalError: local variable 'region_name' referenced before assignment
```

jmunhoz:
@andrewgaul  I found our current boto version not very reliable to test AWS4 in the past. I faced similar issues with regions and/or adding the port in an explicit way (80, 8080 and so on). Both values are used in AWS4 though. It looks our boto version is quite old and it could make sense going with the last boto3 instead of fixing/supporting the current boto version. Not sure though. In the meanwhile, I am using my local DNS to bypass these issues with boto in order to have AWS4 in place ASAP.

To test AWS4 I use one region requiring AWS4 only. In my case I map 's3.eu-central-1.amazonaws.com' to 127.0.0.1. This way I don't depend on boto negotiating AWS2/AWS4 and/or using different parameters. It is useful to develop/test.

With this previous configuration you can run the following command to test AWS4:

$ S3_USE_SIGV4=1 S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests/functional/test_headers.py -a auth_aws4

The previous command would run test cases making sense in aws4_only.

PR #86, opened 2015-11-25
s3tests: Check content is correct on copy multipart

jmunhoz:
Check content is correct on copy multipart


jmunhoz:
@yehudasa this is the missing test case from https://github.com/ceph/s3-tests/pull/75

PR #90, opened 2016-01-04
add a test for mpu atomicity

kahing:
mpu that's aborted should not modify the original object

@andrewgaul


andrewgaul:
:-1: Duplicates `test_atomic_multipart_upload_write` committed in c788b7d0cdd2203da1b2e019fe39dfd50784b4fb.

PR #95, opened 2016-02-08
Added test case to check base 64 encoded metadata is being supported

sivakrishnan47:
Additional check to confirm if base64 encoded metadata is being supported by S3 as well as other storage services


yehudasa:
@sivakrishnan47 sorry for the long turn around for this PR. Looking good, just need a Signed-off-by header in the commits message. Thanks!

Issue #72, opened 2015-08-07
Add conditional get tests


andrewgaul:
Amazon supports conditional gets:

* If-Modified-Since
* If-Unmodified-Since
* If-Match
* If-None-Match

s3-tests should include tests for these.  Note that these are different than the existing and non-standard conditional put tests.  API reference:

https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html

PR #96, opened 2016-02-18
Test copy if-match and if-none-match

gaul:
References #72.

Signed-off-by: Andrew Gaul andrew@gaul.org


andrewgaul:
Rebased and narrowed to copy if-match and if-none-match.

andrewgaul:
@yehudasa bump

gaul:
@alimaredia bump

alimaredia:
@gaul these look great. do you think it would be out of scope for this PR to write similar tests for [upload_part_copy](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.upload_part_copy) and add them to this PR? 

If you don't think so, I'll go ahead and merge this and then write some for `upload_part_copy` myself.

gaul:
I opened #293 to track the suggested tests.

PR #99, opened 2016-03-28
Add fails_on_rgw on multipart copy part tests

jmunhoz:

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -4927,6 +4927,7 @@ def _check_key_content(src, dst):
 @attr(resource='object')
 @attr(method='put')
 @attr(operation='check multipart copies with single small part')
+@attr('fails_on_rgw')
 def test_multipart_copy_small():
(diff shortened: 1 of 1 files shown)


robbat2:
@jmunhoz I have a better variant, that lets the testsuite run the existing test against AWS, and expect the NotImplemented error on rgw, posting shortly.

jmunhoz:
@robbat2 I am not sure if it is good idea checking by different results coming from aws and rgw. The same test logic should pass on aws and rgw in order to be compliant with the s3 interface. The NotImplemented error is just a placeholder to get a clear error while the feature is not in. That's the reason because it has not a test case. I think checking by NotImplemented shouldn't be part of this test case.

I would suggest to keep the fails_on_rgw tag to be consistent with other similar tests in the suite. Another option could be skipping the test although I like the 'fails_on_rgw' option more. What do you think?

robbat2:
https://github.com/ceph/s3-tests/pull/101 is my alternative version of it, just uses fails_on_rgw/fails_on_aws differently.

PR #101, opened 2016-03-30
multipart copy range did not land in Jewel

robbat2:
- RGW tests should trigger NotImplemented for now
- AWS tests should be work for comparision.

See note
https://github.com/ceph/s3-tests/commit/126da37dfd532fc06826d6ccf4510b9fe33e7112#commitcomment-16856581
https://github.com/ceph/s3-tests/pull/99

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -4926,7 +4926,23 @@ def _check_key_content(src, dst):
 
 @attr(resource='object')
 @attr(method='put')
-@attr(operation='check multipart copies with single small part')
+@attr(operation='check multipart copies with single small part - ranges not NotImplemented')
+@attr(assertion='fails')
+@attr('fails_on_aws')
+def test_multipart_copy_small_notimplemented():
+    (src_bucket, src_key) = _create_key_with_random_content('foo')
+    dst_bucket = get_new_bucket()
+    dst_keyname = "mymultipart"
+    size = 1
+    e = assert_raises(boto.exception.S3ResponseError, _multipart_copy, src_bucket.name, src_key.name, dst_bucket, dst_keyname, size)
+    eq(e.status, 501)
+    eq(e.error_code, u'NotImplemented')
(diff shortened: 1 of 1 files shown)


robbat2:
This is an alternative to PR#99

s3tests/functional/test_s3.py
```diff
@@ -4926,7 +4926,23 @@ def _check_key_content(src, dst):
...
+@attr(assertion='fails')
+@attr('fails_on_aws')
+def test_multipart_copy_small_notimplemented():
+    (src_bucket, src_key) = _create_key_with_random_content('foo')
+    dst_bucket = get_new_bucket()
+    dst_keyname = "mymultipart"
+    size = 1
+    e = assert_raises(boto.exception.S3ResponseError, _multipart_copy, src_bucket.name, src_key.name, dst_bucket, dst_keyname, size)
+    eq(e.status, 501)
+    eq(e.error_code, u'NotImplemented')
```
andrewgaul: Why is it so important to test the temporary rgw `NotImplemented` code path?  Other S3 implementations like S3Proxy and akashic use s3-tests and this clutter is just another thing for us to work around.  Why not just annotate with `fails_on_rgw` and be done with it?

robbat2: @andrewgaul what worries me is that I've seen some tests stay annotated with `fails_on_rgw` long after they actually started working on RGW.

To deter a repeat of that, I want something that will explicitly fail and need somebody to update the test when the mp copy range code lands in the next release after Jewel. Didn't realize S3Proxy/akashic were using these as-is, because the other s3-tests consumers I'd seen have diverged heavily.

As an alternative, how can we introduce something that says the present RGW will return NotImplement, without affecting non-RGW non-AWS consumers?

oritwas: I agree with @andrewgaul

robbat2:
ok, closing this then.

PR #102, opened 2016-03-31
[DNM] connection: add a debug option to debug boto calls

theanalyst:
Basically setting debug level via conf file and setting a stream logger
prefixed with 'boto' for more debugging

Signed-off-by: Abhishek Lekshmanan abhishek@suse.com


dachary:
pushed at https://github.com/ceph/s3-tests/tree/wip-boto-debug

PR #106, opened 2016-04-10
s3tests: bucket nuke should also include versioned keys and unfinishe…

robbat2:
…d multipart uploads

### DIFF
--- a/s3tests/common.py
+++ b/s3tests/common.py
@@ -53,6 +53,29 @@ def nuke_bucket(bucket):
                 key.set_canned_acl('private')
                 key.delete()
                 deleted_cnt += 1
+        # Delete old versioned keys
+        deleted_cnt = 1
+        while deleted_cnt:
+            deleted_cnt = 0
+            for key in bucket.list_versions():
+                print 'Cleaning bucket {bucket} key {key} version {verion}'.format(
+                    bucket=bucket,
+                    key=key,
+                    version=key.version_id,
+                    )
+                key.set_canned_acl('private')
+                key.delete()
+                deleted_cnt += 1
+        # Delete all multiparts
+        deleted_cnt = 1
+        while deleted_cnt:
+            deleted_cnt = 0
+            for mp in bucket.list_multipart_uploads():
+                print 'Cleaning bucket {bucket} multipart {mp}'.format(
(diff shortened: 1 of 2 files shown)


yehudasa:
@robbat2 is this one fixing the static website related PR? or is there any other fix in the pipeline?

robbat2:
It's unrelated to staticsites, just improves cleanup to actually clean up more stuff that otherwise gets missed or left for the orphan & garbage collection.

s3tests/common.py
```diff
@@ -53,6 +53,29 @@ def nuke_bucket(bucket):
...
+        deleted_cnt = 1
+        while deleted_cnt:
+            deleted_cnt = 0
+            for key in bucket.list_versions():
+                print 'Cleaning bucket {bucket} key {key} version {verion}'.format(
+                    bucket=bucket,
+                    key=key,
+                    version=key.version_id,
+                    )
+                key.set_canned_acl('private')
```
yehudasa: @robbat2 why do we need to set acls here?

yehudasa: @robbat2 moreover, why not modifying the regular object removal above to do versioned object listing, so that it removes all objects, instead of doing it in two paths?

s3tests/common.py
```diff
@@ -53,6 +53,29 @@ def nuke_bucket(bucket):
...
+                key.delete()
+                deleted_cnt += 1
+        # Delete all multiparts
+        deleted_cnt = 1
+        while deleted_cnt:
+            deleted_cnt = 0
```
yehudasa: @robbat2 what's the purpose of delete_cnt? it's not getting set

s3tests/common.py
robbat2: - the ACL setting is copied from the original block, and seems to be rooted in fixing cases of weird bucket/key ownership from the test suite (the bucket owner gets permission back to the object as a result of changing the ACL).
- If the S3 implementation does not support versioned objects, then `list_versions` is not usable.

s3tests/common.py
robbat2: Nice catch, should have another deleted_cnt += 1 inside the nested loop.

PR #109, opened 2016-04-19
Add test to verify if Asterisk Origin is returned when * is set as a …

wido:
…Allowed Origin.

For example:

The request sends 'Origin: example.origin' while * is allowed as
a origin by the CORS rule.

The Server should then respond with * if no Authorization was send.

Signed-off-by: Wido den Hollander <wido@widodh.nl>


wido:
The tests succeed and I also verified it with Wireshark that the Server indeed responds with the proper headers.

yehudasa:
@wido not crazy about the sleep(3) there, but we do it somewhere else, so for now it's ok. We'll want eventually to have this configurable, or find some other way to do it

PR #111, opened 2016-05-17
s3tests: modify test_get_object_ifmodifiedsince_failed

yehudasa:
Use key's last_modified instead of local time

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -2357,11 +2357,14 @@ def test_get_object_ifmodifiedsince_failed():
     key = bucket.new_key('foo')
     key.set_contents_from_string('bar')
 
-    # Sleep since Amazon returns 200 if the date is in the future:
-    # https://forums.aws.amazon.com/message.jspa?messageID=325930
-    now = time.time()
-    time.sleep(20)
-    after = formatdate(now + 10)
+    for k in bucket.get_all_keys():
+        key = k
+
+    mtime = time.strptime(key.last_modified, '%Y-%m-%dT%H:%M:%S.%fZ')
+
+    after = time.ctime(time.mktime(mtime) + 1)
+
+    time.sleep(1)
 
     e = assert_raises(boto.exception.S3ResponseError, bucket.get_key, 'foo', headers={'If-Modified-Since': after})
     eq(e.status, 304)


yehudasa:
@andrewgaul here's the modified test, how does this one look?

s3tests/functional/test_s3.py
```diff
@@ -2359,9 +2359,13 @@ def test_get_object_ifmodifiedsince_failed():
 
     # Sleep since Amazon returns 200 if the date is in the future:
     # https://forums.aws.amazon.com/message.jspa?messageID=325930
```
andrewgaul: Stale comment.

s3tests/functional/test_s3.py
```diff
@@ -2359,9 +2359,13 @@ def test_get_object_ifmodifiedsince_failed():
...
-    now = time.time()
-    time.sleep(20)
-    after = formatdate(now + 10)
+
+    for k in bucket.get_all_keys():
+        key = k
+
+    mtime = time.strptime(key.last_modified, '%Y-%m-%dT%H:%M:%S.%fZ')
+
+    after = time.ctime(time.mktime(mtime) + 1)
```
andrewgaul: Shouldn't you sleep for 1 second to avoid racing against the clock when latency is small?

s3tests/functional/test_s3.py
```diff
@@ -2359,9 +2359,13 @@ def test_get_object_ifmodifiedsince_failed():
 
     # Sleep since Amazon returns 200 if the date is in the future:
     # https://forums.aws.amazon.com/message.jspa?messageID=325930
-    now = time.time()
-    time.sleep(20)
-    after = formatdate(now + 10)
+
+    for k in bucket.get_all_keys():
+        key = k
```
andrewgaul: Better to call `bucket.get_key('foo')`?

s3tests/functional/test_s3.py
yehudasa: right. Had that, but remove that line accidentally.

s3tests/functional/test_s3.py
yehudasa: ok

s3tests/functional/test_s3.py
yehudasa: @andrewgaul tried that but it returns a different time format that is harder to work with

andrewgaul:
:+1:

PR #114, opened 2016-06-02
DNM: Tests for server side encryption, S3 SSE-C mode.

aclamk:
These tests are intended to verify encryption in radosgw.


cbodley:
@aclamk can we close this one?

aclamk:
@cbodley It is superseded by encryption-s3-sse-kms. Can be closed.

PR #119, opened 2016-07-22
Add range request to an empty object

oritwas:
Signed-off-by: Orit Wasserman <owasserm@redhat.com>


cbodley:
@oritwas looks good, and i was able to see it work when testing https://github.com/ceph/ceph/pull/10207. but it says i'm not authorized to merge

oritwas:
I guess only @yehudasa can merge it

PR #124, opened 2016-09-09
Complete multi-part upload during dry-run

tipabu:
While it may be valuable to verify that deleting a non-existent key is OK, that doesn't seem to be the point of this test.


andrewgaul:
👍

yehudasa:
@tipabu can you sign off this PR?

tipabu:
@yehudasa done. Sorry, I didn't see any contribution guidelines here, so I wasn't sure what the protocol was.

PR #126, opened 2016-09-22
S3 lifecycle tests

dang:
Test S3 lifecycle expiration.  This requires the test harnass code in
RGW to make sure it doesn't take days so expire anything.

Signed-off-by: Daniel Gryniewicz <dang@redhat.com>

Needs the same PR for ceph-qa-suite

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -2,6 +2,7 @@ from cStringIO import StringIO
 import boto.exception
 import boto.s3.connection
 import boto.s3.acl
+import boto.s3.lifecycle
 import bunch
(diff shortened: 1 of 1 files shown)


yehudasa:
@dang since this depends on some rgw code, and this suite can run against other backends (non-rgw), we need to skip these tests when not dealing with rgw

yehudasa:
@dang one option is to have a new config option here that will specify whether backend supports accelerated time for lifecycle

dang:
All of these have the "lifecycle" attribute.  You can just skip the tests if they're not supported.  Or am I missing something else?

PR #135, opened 2016-10-31
Fix If-Match test

jkorvin:
According to https://tools.ietf.org/html/rfc7232, If-Match field should looks like QUOTES ETag QUOTES.


yehudasa:
@jkorvin can you add Signed-off-by header to the commit?

jkorvin:
@yehudasa done

PR #136, opened 2016-11-09
Fix 'If-Modified-Since' format string

jkorvin:
According to https://tools.ietf.org/html/rfc7232
header 'If-Modified-Since' should be a correct HTTP-date.
For example: `Tue, 15 Nov 1994 12:45:26 GMT`;
not `1994-11-15T12:45:26.000Z`.


yehudasa:
@jkorvin can you add Signed-off-by to this commit?

jkorvin:
@yehudasa done

PR #150, opened 2017-02-04
remove python2 pip

vasukulkarni:
python2-pip doesn't exist in rhel and python-virtualenv ships pip in its package.

Signed-off-by: Vasu Kulkarni <vasu@redhat.com>


vasukulkarni:
http://pulpito.ceph.com/vasu-2017-02-04_01:45:36-smoke-kraken---basic-vps/

PR #118, opened 2016-07-11
DNM: Encryption s3 sse kms

aclamk:
This are codes that test features implemented by: 
https://github.com/aclamk/ceph/tree/wip-rgw-s3-crypto-sse-kms
If above is not merged, most of tests will fail.

PR #154, opened 2017-03-13
functional tests for server side encryption

cbodley:
this is the same as https://github.com/ceph/s3-tests/pull/118, except it's been rebased and includes a fix for some HEAD requests that now require a customer key header

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -7589,6 +7589,29 @@ def test_encrypted_transfer_13b():
 
 
 @attr(resource='object')
+@attr(method='head')
+@attr(operation='Test SSE-C encrypted does perform head properly')
+@attr(assertion='success')
+@attr('encryption')
+def test_encryption_sse_c_method_head():
+    bucket = get_new_bucket()
+    sse_client_headers = {
+        'x-amz-server-side-encryption-customer-algorithm': 'AES256',
+        'x-amz-server-side-encryption-customer-key': 'pO3upElrwuEXSoFwCfnZPdSsmt/xWeFa0N9KgDijwVs=',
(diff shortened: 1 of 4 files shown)


s3tests/functional/test_s3.py
```diff
@@ -7533,3 +7533,737 @@ def test_lifecycle_rules_conflicted():
...
+
+@attr(resource='object')
+@attr(method='put')
+@attr(operation='write encrypted with SSE-C but read with other key')
+@attr(assertion='operation fails')
+@attr('encryption')
+def test_encryption_sse_c_other_key():
+    bucket = get_new_bucket()
+    sse_client_headers_A = {
+        'x-amz-server-side-encryption-customer-algorithm': 'AES256',
```
cbodley: i don't see much value in making it configurable. the `kms_keyid` is different, because it requires you to set up a key in barbican and plug in its id. for these sse-c requests, the client can pass in any key and expect it to work (as long as it reads and writes with the same key)

the only dependency we have on the value of this key, is when the qa suite (https://github.com/ceph/ceph/pull/13597) greps radosgw logs to make sure we're not leaking encryption keys. so either we hardcode them both (as currently), or we make both configurable and have the rgw task make sure they match. i'm not sure that's worth the effort, but i'll defer to your judgement here

cbodley:
@oritwas thanks for review, i removed the line that changed `mymultipart`

s3tests/functional/test_s3.py
oritwas: lets leave as is

cbodley:
thanks @oritwas!

@yehudasa @mattbenjamin could one of you please merge?

PR #158, opened 2017-04-12
Test case for non-current expiration，multipart upload expiration and delete marker expiration

zhangsw:
These cases are for non-current version object expiration and multipart upload expiration in s3 lifecycle. Related features are in https://github.com/ceph/ceph/pull/13622 and https://github.com/ceph/ceph/pull/13385 .
Because boto doesn't support those features, so the xml body is generated manually.

Fixes: http://tracker.ceph.com/issues/19519

Signed-off-by: Zhang Shaowen <zhangshaowen@cmss.chinamobile.com>

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -7533,6 +7533,175 @@ def test_lifecycle_rules_conflicted():
     eq(e.status, 400)
     eq(e.error_code, 'InvalidRequest')
 
+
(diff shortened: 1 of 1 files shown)


zhangsw:
@yehudasa @dang Please help review this, thanks~

zhangsw:
Add the support for delete marker expiration in https://github.com/ceph/ceph/pull/14703

PR #162, opened 2017-04-28
[WIP] support absolute urls as a calling format

theanalyst:
This is a patch to support Absolute Calling Format wherein one passes the full url as an argument while making requests (ie. GET http://foobar.com/bar instead of just GET /bar)


theanalyst:
this still breaks wherever calling_format=subdomain fails as we are not passing the info. about path style for a few of the `_make_bucket_request` calls

theanalyst:
Updated the code to pass by checking the type in `_make_request`

theanalyst:
calling_format=absolute depends on https://github.com/ceph/ceph/pull/12861

PR #166, opened 2017-06-13
clear key.version_id now that GET requests return x-amz-version-id header

cbodley:
https://github.com/ceph/ceph/pull/14117 added the `x-amz-version-id` header to GetObj requests. boto's [Key.get_contents_as_string()](http://boto.cloudhackers.com/en/latest/ref/s3.html#boto.s3.key.Key.get_contents_as_string) will set `Key.version_id` based on this header. this is causing `test_versioning_obj_plain_null_version_overwrite()` to issue a `get_contents_as_string()` against the version_id it had just deleted, failing with `S3ResponseError: 404 Not Found`


cbodley:
retargeted from ceph-master branch to master. will cherry-pick to ceph-master after merge here

PR #172, opened 2017-06-23
Tag most of the the Browser Upload tests as AWSv2 specific.

rzarzynski:
Altough most of the Browser Upload tests are truly specific to AWSv2, they were being executed also in AWSv4 runs causing many false-positives. The patch tag them appropriately and prevents from being executed when `S3_USE_SIGV4=1` is specified.

Signed-off-by: Radoslaw Zarzynski <rzarzyns@redhat.com>

### DIFF
--- a/s3tests/functional/test_headers.py
+++ b/s3tests/functional/test_headers.py
@@ -11,7 +11,6 @@ import random
 import string
 import socket
 import ssl
-import os
 import re
 
 from urlparse import urlparse
@@ -20,9 +19,10 @@ from boto.s3.connection import S3Connection
 
 from nose.tools import eq_ as eq
 from nose.plugins.attrib import attr
-from nose.plugins.skip import SkipTest
 
 from .utils import assert_raises
+from .utils import check_aws2_support
+from .utils import check_aws4_support
 import AnonymousAuth
 
 from email.header import decode_header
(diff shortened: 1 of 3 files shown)


yehudasa:
@rzarzynski I think we're mixing here two different issues. We should be able to do browser upload in conjuction with aws4. Preferably these should be the same tests doing it. So I think the correct fix for these specific tests that fail would be to adapt the tests to support both aws2 and aws4, and not limit these to aws2.

yehudasa:
@rzarzynski that been said, there probably are tests that we want to limit to aws2, and your cleanup looks good,.

rzarzynski:
> We should be able to do browser upload in conjuction with aws4.

Fully agreed. We really need to have such tests also for AWSv4.

> Preferably these should be the same tests doing it.

Personally I perceive this as a detail. We could cover both v2 and v4 in a single test (with internal branch) or, alternatively, have two different tests named appropriately (e.g. `test_post_object_authenticated_request_bad_access_key_aws2` and `test_post_object_authenticated_request_bad_access_key_aws4`). The first approach looks compliant with DRY principle.

> I think the correct fix for these specific tests that fail would be to adapt the tests to support both aws2 and aws4, and not limit these to aws2.

Absolutely. Skipping these tests is only a makeshift solution to be more explicit. Currently, they are named in a way that doesn't reflect that only AWSv2 is verified.

yehudasa:
@rzarzynski I'm afraid that if we're going to skip these tests now, we'll never get to fix the tests.

rzarzynski:
@yehudasa: yeah, that's possible, unfortunately. Looks like we need to judge what's the lesser evil for now. :-(

I can take fixing them but, preferably, would finish the Keystone/Tempest integration stuff first.

PR #179, opened 2017-08-02
Add test for invalid copy part range

andrewgaul:
References kahing/goofys#212.

Signed-off-by: Andrew Gaul <andrew@gaul.org>


cbodley:
@yehudasa should this be tested upstream as part of ceph-master? it's currently failing against radosgw, because we return 416 InvalidRange instead of 400

yehudasa:
@cbodley we can modify this one to check either

PR #181, opened 2017-08-14
lc: set attribute as lifecycle_time for lc tests using timing ops

theanalyst:
Allows us to temporarily filter out tests based on timing attribute
which currently fail in the teuthology runs

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -7534,6 +7534,7 @@ def test_lifecycle_get():
 @attr(method='put')
 @attr(operation='test lifecycle expiration')
 @attr('lifecycle')
+@attr('lifecycle_expiration')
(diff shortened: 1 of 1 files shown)


s3tests/functional/test_s3.py
```diff
@@ -7533,7 +7533,7 @@ def test_lifecycle_get():
 @attr(resource='bucket')
 @attr(method='put')
 @attr(operation='test lifecycle expiration')
-@attr('lifecycle')
+@attr('lifecycle_timing')
```
cbodley: maybe keep the `lifecycle` tag as it is, and add another one like `expiration` or `lifecycle_expiration`?

theanalyst: sure makes sense, since nose will filter the attribute correctly anyway

PR #187, opened 2017-09-20
add the LC fixes to ceph-master

theanalyst:
cherry-picking the LC fixes to ceph-master, we cant forward master to ceph-master yet because it looks like the test for invalid copy range doesn't pass yet


theanalyst:
@cbodley @yehudasa is this good to go?

PR #188, opened 2017-09-20
add tests for lifecycle with tagging

theanalyst:
test LC with tagging, we have XML tests that validate that we have a
valid XML that various conditions in the Filter are satisfied (ie.
multiple conditions should have an AND tag, both prefix and multiple
object tags are supported) also one functional test which would upload a
bunch of objects, an object with tags during put object itself (and
others with explicit put_object_tagging calls to validate the mtimes are
ok), and validate that we only delete objects which include all the tags
specified in LC policy


theanalyst:
@dang @yehudasa can we merge this, the corresponding feature was already merged in master with https://github.com/ceph/ceph/pull/17305

PR #191, opened 2017-10-13
test bucket policy with conditionals 

theanalyst:
Adding tests for bucket policies with conditionals
Tests added so far:
- Get Object (with ExistingObjectTag)
- Put Object Tagging
- Get Object Tagging 
- Put Object
- Put Object ACL
- Get Object ACL

- CreateBucket
- ListBucket (with prefix/max-keys/delimiter conditionals)

### DIFF
--- a/s3tests/functional/policy.py
+++ b/s3tests/functional/policy.py
@@ -1,0 +1,38 @@
+import json
+
+class Statement(object):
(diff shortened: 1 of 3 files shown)


theanalyst:
Changelog: updated the helper function to reuse the existing Policy classes

PR #193, opened 2017-10-30
[DNM]test for https://github.com/ceph/ceph/pull/17934

kmroz:
http://tracker.ceph.com/issues/21939


cbodley:
@kmroz should we wait for https://github.com/ceph/ceph/pull/18569 to merge? don't want to confuse people with luminous failures from this new test

kmroz:
@cbodley - I should have marked this dnm, sorry about that.

cbodley:
it looks like this commit is already present in https://github.com/ceph/s3-tests/commits/ceph-luminous

PR #199, opened 2017-11-16
Add test case for getting version-id when upload objects  bucket

cfanz:
This test case is used for testing whether a version-id header is returned when upload files.

Related with [ceph #18935](https://github.com/ceph/ceph/pull/18935)

Signed-off-by: Xinying Song <songxinying@cloudin.cn>

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -34,6 +34,7 @@ from httplib import HTTPConnection, HTTPSConnection
 from urlparse import urlparse
 
 from nose.tools import eq_ as eq
+from nose.tools import assert_is_none
 from nose.plugins.attrib import attr
 from nose.plugins.skip import SkipTest
 from boto.s3.tagging import TagSet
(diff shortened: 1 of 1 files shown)


s3tests/functional/test_s3.py
```diff
@@ -9315,3 +9316,67 @@ def test_delete_tags_obj_public():
...
+    key_name = 'baz'
+    key = bucket.new_key(key_name)
+    key.set_contents_from_string(key_name)
+    assert_is_none(key.version_id)
+
+    # for versioning-suspended-bucket, no version-id should return.
+    bucket = get_new_bucket()
+    key_name = 'foo'
+    key = bucket.new_key(key_name)
+    key.set_contents_from_string(key_name)
```
cbodley: it looks like this last part is missing the call to suspend versioning and the check for `key.version_id`

s3tests/functional/test_s3.py
```diff
@@ -9315,3 +9316,67 @@ def test_delete_tags_obj_public():
...
+    bucket = get_new_bucket()
+    check_configure_versioning_retry(bucket, True, "Enabled")
+    key_name = 'bar'
+    key = bucket.new_key(key_name)
+    key.set_contents_from_string(key_name)
+    # list_verions will return an non-empty version-id
+    li = bucket.list_versions()
+    for k in li:
+        eq(key.version_id, k.version_id)
+    return
```
cbodley: need to remove the return here

cfanz: @cbodley  Oops! Sorry for my carelessness. Have fixed it.

cbodley:
cherry-picked to ceph-master

PR #208, opened 2018-01-31
add list multipart upload test

liaoxin01:
Signed-off-by: liaoxin <liaoxinbit@gmail.com>


liaoxin01:
This pr is prepared for [ceph/ceph#20158](https://github.com/ceph/ceph/pull/20158)

s3tests/functional/test_s3.py
```diff
@@ -5709,6 +5709,83 @@ def test_list_multipart_upload():
     upload2.cancel_upload()
     upload3.cancel_upload()
 
+@attr(resource='bucket')
+@attr(method='put')
+@attr(operation='list multi-part uploads')
+def test_list_multipart_upload_many():
+    bucket = get_new_bucket()
+    mb = 1024 * 1024
+    (upload1, data) = _multipart_upload(bucket, "foo", 5 * mb)
```
gaul: parenthesis not needed.

s3tests/functional/test_s3.py
```diff
@@ -5709,6 +5709,83 @@ def test_list_multipart_upload():
...
+    (upload4, data) = _multipart_upload(bucket, "sample.jpg", 5 * mb)
+
+    l = bucket.get_all_multipart_uploads(prefix='photos')
+    l = list(l)
+    eq(len(l), 2)
+
+    index = dict([("photos/2006/January/sample.jpg", 1),
+                  ("photos/2006/February/sample.jpg", 1)])
+    for upload in l:
+        index[upload.key_name] -= 1;
```
gaul: use `set` instead?

s3tests/functional/test_s3.py
```diff
@@ -5709,6 +5709,83 @@ def test_list_multipart_upload():
...
+    mb = 1024 * 1024
+    (upload1, data) = _multipart_upload(bucket, "photos/2006/January/sample.jpg", 5 * mb)
+    (upload2, data) = _multipart_upload(bucket, "photos/2006/February/sample.jpg", 5 * mb)
+    (upload3, data) = _multipart_upload(bucket, "sample.jpg", 5 * mb)
+    (upload4, data) = _multipart_upload(bucket, "sample.wmv", 5 * mb)
+
+    l = bucket.get_all_multipart_uploads(delimiter='/')
+    eq(len(l), 3)
+ 
+    prefix = getattr(l[-1], "CommonPrefixes.Prefix")
```
gaul: Doe `l[-1].prefix` not work?

liaoxin01: It does't work. I can't find a better way to get the prefix .

s3tests/functional/test_s3.py
liaoxin01: It's ok to use set here. Because the  name of two multi_uploads may be same, it will list all of them. So use dict may be better.  I wrote this test for reference to the case of "test_list_multipart_upload".

PR #210, opened 2018-02-07
boto3: Foundation laid for boto3 tests

alimaredia:
Added sample config file for boto3 and vstart.sh

Modified setup.py, requirements.txt, and README

Added an rgw_interactive.py to use interactively
with vstart.sh and python -i

Ported 250+ tests over to boto3 from functional/test_s3.py

Signed-off-by: Ali Maredia <amaredia@redhat.com>


s3tests_boto3/functional/__init__.py
```diff
@@ -0,0 +1,355 @@
...
+
+def get_buckets_list(client=None, prefix=None):
+    if client == None:
+        client = get_client()
+    if prefix == None:
+        prefix = get_prefix()
+    response = client.list_buckets()
+    bucket_dicts = response['Buckets']
+    buckets_list = []
+    if bucket_dicts != []:
```
theanalyst: We can probably avoid the empty list check since the for loop will exit for empty lists anyway

s3tests_boto3/functional/test_s3.py
```diff
@@ -0,0 +1,6895 @@
...
+
+    r = requests.post(url, files = payload)
+    eq(r.status_code, 400)
+
+@attr(resource='object')
+@attr(method='post')
+@attr(operation='authenticated browser based upload via POST request')
+@attr(assertion='fails with missing key error')
+def test_post_object_no_key_specified():
+
```
theanalyst: we can probably move some of the post tests with http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.generate_presigned_post (okay for now)

alimaredia: good call, I'll add that to the todo list.

s3tests_boto3/functional/__init__.py
alimaredia: I'm going to force-push a new version of this soon, and this change will be made.

PR #211, opened 2018-02-09
copy object: return 'x-amz-copy-source-version-id & 'x-amz-version-id'

fangyuxiangGL:


fangyuxiangGL:
@mattbenjamin As we talked in https://github.com/ceph/ceph/pull/20348, you could verify the case, thanks.

PR #213, opened 2018-03-02
splunk s3-compatibility checking tool v1

hijiahanwang:
Will add the readme file w/ customer instruction after review. Also need to move entire thing to splunk's public github account later after going thru necessary open-sourcing steps.


hijiahanwang:
NVM, create the PR at wrong place :)

PR #217, opened 2018-03-27
unpin the requests library version

cbodley:
the old requests library doesn't support recent versions of tls


cbodley:
ping @alimaredia

PR #223, opened 2018-04-26
Abort any multipart uploads so the bucket can be deleted

GrandPubba:


KhasanovBI:
Is this behavior described in the documentation? Can you provide link please if yes?

gaul:
Could you add a test for deleting a bucket which has an in-progress MPU?

PR #224, opened 2018-05-09
Fix several compatibility issues

zhouk4:
1) Fix the reason string returned by nginx
2) Re-create the same bucket will fail with BucketAlreadyOwnedByYou.
Fix test_bucket_recreate_not_overriding by referring to another ceph test case test_bucket_create_exists
3) The order of XML returned by lifecycle get is random.
Fix it by referring to another ceph test case test_lifecycle_get_no_id
4) Fix not supported pre-signed URL using SSE with specific customer-provided encryption keys.


gaul:
2 related to #10.

PR #229, opened 2018-05-31
Fix some dependencies in s3tests

zhenningz-git:
1. Update minimum requests version to 2.10.0
2. Remove dependency of the old ordereddict and use default collections.OrderedDict


cbodley:
@alimaredia could you please help test this in teuthology to make sure it works on all our distros?

PR #246, opened 2018-11-16
Force python2 virtualenv

gaul:
python defaults to 2.7 on Fedora 29 yet virtualenv chooses 3.7
instead.  Force python2 to work around this.


gaul:
Superseded by #264.

PR #248, opened 2018-11-30
Update outdated bucket naming tests

tchutw:
It seems like 'fails_on_aws' was intended to mark cases when AWS didn't enforce their own documentation, but when these tests began to fail when AWS changed the bucket naming rules to enforce DNS-compliance, it seems like they were marked as 'fails_on_aws' instead of properly changing the tests.

The bucket name enforcement rules in us-east-1 were updated to match those for other regions (as of March 2018), making these tests now over a half-year out of date for all regions.

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -3276,27 +3276,6 @@ def test_bucket_create_naming_bad_long_254():
 def test_bucket_create_naming_bad_long_255():
     _test_bucket_create_naming_bad_long(255)
 
-'''
-# Breaks DNS with SubdomainCallingFormat
-@attr('fails_with_subdomain')
-@attr(resource='bucket')
-@attr(method='get')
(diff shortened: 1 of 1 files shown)


s3tests/functional/test_s3.py
```diff
@@ -3272,9 +3273,10 @@ def test_bucket_create_naming_good_long_254():
 @attr(method='put')
 @attr(operation='create w/255 byte name')
 @attr(assertion='fails with subdomain')
-def test_bucket_create_naming_good_long_255():
-    _test_bucket_create_naming_good_long(255)
+def test_bucket_create_naming_bad_long_255():
+    _test_bucket_create_naming_bad_long(255)
 
+'''
```
gaul: If this test is no longer valid can you remove it?

tchutw: Ok. Done in a 2nd commit.

PR #249, opened 2018-12-11
test versioned copy with urlencoded keys

cbodley:
test case for https://github.com/ceph/ceph/pull/25498


cbodley:
@alimaredia i rewrote this in boto3, can you please review?

alimaredia:
Test is passing on master: http://qa-proxy.ceph.com/teuthology/amaredia-2019-06-04_21:38:09-rgw:verify-master-distro-basic-smithi/4005101/teuthology.log

PR #252, opened 2019-01-21
rgw: boto3: fix ssl endpoints

alimaredia:

### DIFF
--- a/s3tests_boto3/functional/__init__.py
+++ b/s3tests_boto3/functional/__init__.py
@@ -155,7 +155,7 @@ def setup():
     # vars from the DEFAULT section
     config.default_host = defaults.get("host")
     config.default_port = int(defaults.get("port"))
-    config.default_is_secure = defaults.get("is_secure")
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/__init__.py
```diff
@@ -155,7 +155,7 @@ def setup():
     # vars from the DEFAULT section
     config.default_host = defaults.get("host")
     config.default_port = int(defaults.get("port"))
-    config.default_is_secure = defaults.get("is_secure")
+    config.default_is_secure = defaults.get("is_secure") in [ 'true', 'yes', '1' ]
```
cbodley: s3tests/functional/\_\_init\_\_.py uses `cfg.getboolean(section, 'is_secure')`, which handles these cases. can you try `defaults.getboolean("is_secure")` here?

cbodley: oh, it's yehuda's branch. if it's passing tests this way, let's get it in as is

PR #254, opened 2019-01-31
added listobjects v2

albIN7:


s3tests_boto3/functional/test_s3.py
```diff
@@ -5102,6 +5102,38 @@ def test_bucket_recreate_not_overriding():
...
+
+@attr(resource='bucket')
+@attr(method='get')
+@attr(operation='list all objects in a bucket using list_objects_v2')
+@attr(assertion='returns all expected objects')
+
+def test_listobjects_v2():
+    client = get_client()
+    bucket_names = []
+    for i in xrange(5):
```
mattbenjamin: a little odd juggling names--but looks valid, must be similar logic for v1

albIN7: Yes, similiar logic is followed in v1

s3tests_boto3/functional/test_s3.py
```diff
@@ -5102,6 +5102,42 @@ def test_bucket_recreate_not_overriding():
...
+
+    buckets_list = get_buckets_list()
+
+    for name in bucket_names:
+        if name in buckets_list:
+            response1 = client.list_objects_v2(Bucket=name)
+            print response1
+        else:
+            raise RuntimeError("S3 implementation's GET on Service did not return bucket we created: %r", bucket.name)
+
```
gaul: Shouldn't this test `continuation-token`, `start-after`, and `fetch-owner`?

albIN7: These are optional parameters. We will add more boto scripts that have different requirements that will test all these parameters.

albIN7:
https://github.com/ceph/s3-tests/pull/257

PR #257, opened 2019-02-14
listobjectsv2 with more paramters

albIN7:


albIN7:
@alimaredia

albIN7:
@alimaredia @cbodley 
Added every parameters. Tests are OK.

s3tests_boto3/functional/test_s3.py
```diff
@@ -5102,6 +5102,107 @@ def test_bucket_recreate_not_overriding():
...
+
+    response = client.list_buckets()
+    bucket_dicts = response['Buckets']
+    buckets_list = []
+
+    buckets_list = get_buckets_list()
+
+    for name in bucket_names:
+        if name in buckets_list:
+            response1 = client.list_objects_v2(Bucket=name)
```
alimaredia: @albIN7 the response that this omits is already a dict. You can look for values in the dict like `KeyCount` and use `assert_raises()` to to check for the KeyCount value without raising your own ValueError. This same principle applies to the other tests you have as well.

s3tests_boto3/functional/test_s3.py
```diff
@@ -5102,6 +5102,107 @@ def test_bucket_recreate_not_overriding():
...
+            d = json.loads(response1)
+            if d['KeyCount'] != 5:
+            	raise ValueError
+        else:
```
alimaredia: There's no need for a `RuntimeError `. It's safe to assume in this test that buckets can be created properly.

PR #261, opened 2019-03-18
boto{2,3}: rm functional.test_s3.test_lifecycle_rules_conflict

mattbenjamin:
The assumption that there may be only one rule per prefix has
been removed.  The rules specifically tested here--without tag
filters--do overlap but are not in conflict.

I am proposing to remove altogether rather than writing new
deconfliction logic.


mattbenjamin:
@cbodley updated, please check?

PR #263, opened 2019-03-22
sse: add sse-c test with unaligned multipart parts

cbodley:
Fixes: http://tracker.ceph.com/issues/38700

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -8527,13 +8527,12 @@ def _check_content_using_range_enc(client, bucket_name, key, data, step, enc_hea
 @attr(operation='complete multi-part upload')
 @attr(assertion='successful')
 @attr('encryption')
-@attr('fails_on_aws') # allow-unordered is a non-standard extension
 def test_encryption_sse_c_multipart_upload():
     bucket_name = get_new_bucket()
     client = get_client()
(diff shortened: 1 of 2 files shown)


s3tests_boto3/functional/test_s3.py
```diff
@@ -8571,6 +8571,52 @@ def test_encryption_sse_c_multipart_upload():
...
+@attr(operation='complete multi-part upload with unaligned part size')
+@attr(assertion='successful')
+@attr('encryption')
+def test_encryption_sse_c_multipart_upload():
+    bucket_name = get_new_bucket()
+    client = get_client()
+    key = "multipart_enc"
+    content_type = 'text/plain'
+    objlen = 30 * 1024 * 1024
+    partlen = 1 + 5 * 1024 * 1024
```
theanalyst: one more case to add would be to set partlen as just 5MiB and objlen as say something with an  offby1 , say 30MiB+ 10 which should catch the case of the last unaligned part.

cbodley: i now have the original multipart tests using objlen=30M+1 partlen=5M, and the new unaligned tests using objlen=30M partlen=5M+1

cbodley:
now lives in https://github.com/ceph/s3-tests/pull/266

PR #264, opened 2019-03-22
🐍 Bruising the serpent's head! 🐍

adamemerson:
Away with you, Python 2!


theanalyst:
Are we killing py2 completely or do we need a 2-3 compatible mode using something like six for some of the items?

s3tests/functional/__init__.py
```diff
@@ -1,21 +1,20 @@
-from __future__ import print_function
 import sys
-import ConfigParser
+import configparser
```
theanalyst: these can work in 2 and 3 if we use something like from six.moves import configparser

s3tests/functional/__init__.py
```diff
@@ -1,21 +1,20 @@
...
-import bunch
+import munch
 import itertools
 import os
 import random
 import string
-from httplib import HTTPConnection, HTTPSConnection
-from urlparse import urlparse
+from http.client import HTTPConnection, HTTPSConnection
+from urllib.parse import urlparse
```
theanalyst: similarly six.moves.urllib.parse import urlparse

s3tests/functional/test_headers.py
```diff
@@ -1,10 +1,9 @@
-from cStringIO import StringIO
+from io import StringIO
```
theanalyst: six.stringio

s3tests/functional/test_headers.py
```diff
@@ -24,7 +23,7 @@
 from nose.plugins.skip import SkipTest
 
 from .utils import assert_raises
-import AnonymousAuth
+from . import AnonymousAuth
```
theanalyst: for py2 probably `from __future__ import AbsoluteImport` should turn this behaviour on? (not so sure myself)

theanalyst: and similarly `from __future__ import print_funciton`

requirements.txt
```diff
@@ -2,12 +2,11 @@ PyYAML
...
 boto >=2.6.0
 boto3 >=1.0.0
-bunch >=1.0.0
+munch >=2.0.0
 # 0.14 switches to libev, that means bootstrap needs to change too
 gevent >=1.0
 isodate >=0.4.4
 requests >=0.14.0
 pytz >=2011k
-ordereddict
```
theanalyst: good catch, I think py 2.6+ supported this already

adamemerson:
@theanalyst I was thinking we don't need the extra complexity and bother of Python2 compatibility at this point since it's only got eight months of life left in it.

s3tests/functional/__init__.py
```diff
@@ -110,12 +109,12 @@ def nuke_prefixed_buckets_on_conn(prefix, name, conn):
 def nuke_prefixed_buckets(prefix):
     # If no regions are specified, use the simple method
     if targets.main.master == None:
-        for name, conn in s3.items():
+        for name, conn in list(s3.items()):
```
theanalyst: Does this work without the list(s3.items()) call, i assume dict.items() generator should be iterable without needing to realize the list?

s3tests/functional/test_s3_website.py
```diff
@@ -183,7 +183,7 @@ def __website_expected_reponse_status(res, status, reason):
 
 def _website_expected_default_html(**kwargs):
     fields = []
-    for k in kwargs.keys():
+    for k in list(kwargs.keys()):
```
theanalyst: same argument as above? can this work without realizing the list

s3tests/functional/utils.py
```diff
@@ -42,7 +42,7 @@ def generate_random(size, part_size=5*1024*1024):
 # syncs all the regions except for the one passed in
 def region_sync_meta(targets, region):
 
-    for (k, r) in targets.iteritems():
+    for (k, r) in targets.items():
```
theanalyst: probably the tuple packing can be avoided here, since we don't seem to use the tuple anyway, though mostly a nit.

PR #265, opened 2019-03-25
Add multipart tests with range header 

theanalyst:


theanalyst:
need to add boto2 tests

theanalyst:
this depends on https://github.com/ceph/s3-tests/pull/263 atm, as it makes sense having all the encryption tests together. I can push to that pr as well.

cbodley:
note that the existing multipart test has some calls to `_check_content_using_range_enc()` that does do these range requests and compare the expected data. it might be easier to adapt what's there, so it can test these interesting ranges without having to loop over the entire object?

theanalyst:
makes sense, I'll modify that function to do a few more ranges.

theanalyst:
created #266 which builds atop of your pr and just adds these part size ranges as well to the tests

PR #271, opened 2019-04-05
skip bucket location tests if no api_name is configured

cbodley:


cbodley:
ping @alimaredia can this merge?

PR #276, opened 2019-05-19
Explicitly reference Fedora Python 2 dependency

gaul:
Previously the aliases would install but subsequent runs would not
find the exact package name.  This caused yum to unnecessarily re-run.

Signed-off-by: Andrew Gaul <andrew@gaul.org>


gaul:
@alimaredia Please review.

PR #283, opened 2019-06-21
Fix create_bucket tests as per new bucket naming conventions

soumyakoduri:
As per amazon s3 spec -
https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html

The s3 bucket names should not contain upper case letters or underscore.
Name cannot end with dash or have consecutive periods, or dashes adjacent to periods
Name length shouldn't exceed 63 characters.

These rules are being enforced via
- https://github.com/ceph/ceph/pull/26787

This patch is to update the respective testcases as well.

Signed-off-by: Soumya Koduri <skoduri@redhat.com>


soumyakoduri:
@cbodley @alimaredia .. kindly review the changes.

PR #286, opened 2019-06-28
Add a case for lifecycle expiration on versining enabled bucket.

zhangsw:
This is for https://github.com/ceph/ceph/pull/28587
Signed-off-by: zhang Shaowen <zhangshaowen@cmss.chinamobile.com>


cbodley:
thanks @zhangsw - i'm seeing this test pass against master, though, so i don't think it really exposes the issue with duplicate delete markers

zhangsw:
@cbodley I've updated the case. Now it works well...

cbodley:
@zhangsw excellent, thank you. :+1: with your fix in https://github.com/ceph/ceph/pull/28587, i see it passing. without the fix, it's catching the failure with:
```
    eq(len(delete_markers), 1)                                                                                                                                                                      
AssertionError: 2 != 1
```

PR #287, opened 2019-07-02
make sure that virtualenv is python2 regardless of host

yuvalif:
when running on fedora29 the virtualenv was python3. this didn't work since the tests are pyhton2.

Signed-off-by: Yuval Lifshitz <yuvalif@yahoo.com>


gaul:
Related to #276?

PR #289, opened 2019-07-05
rgw: az: Add archive zone test cases

jmunhoz:
Test cases for the archive sync module feature. It requires to set up
one configuration with two available zones. One of those zones is the
archive zone (az):

http://docs.ceph.com/docs/nautilus/radosgw/archive-sync-module/

To run the test cases you need to configure 'host_az' and 'port_az' in
your s3test conf (see s3tests.conf.SAMPLE) and set up RGW_S3_USE_AZ=1 in
the environment. Use 'host' and 'port' to configure the non archive
zone.

If RGW_S3_USE_AZ is not available in the environment the test cases will
be skipped. It is the default behaviour.

To run the archive zone test cases directly:

$ RGW_S3_USE_AZ=1 S3TEST_CONF=./s3tests.conf \
./virtualenv/bin/nosetests -v -a 'archive-zone' s3tests_boto3.functional

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -40,6 +40,7 @@ from .policy import Policy, Statement, make_json_policy
 
 from . import (
     get_client,
+    get_az_client,
     get_prefix,
     get_unauthenticated_client,
     get_bad_auth_client,
@@ -6677,14 +6678,18 @@ def test_multipart_upload():
     _check_content_using_range(key, bucket_name, data, 1000000)
     _check_content_using_range(key, bucket_name, data, 10000000)
 
-def check_versioning(bucket_name, status):
-    client = get_client()
+def check_versioning(bucket_name, status, client=None):
+    if client == None:
+        client = get_client()
 
     try:
         response = client.get_bucket_versioning(Bucket=bucket_name)
         eq(response['Status'], status)
     except KeyError:
         eq(status, None)
+
+def check_az_versioning(bucket_name, status):
+    check_versioning(bucket_name, status, get_az_client())
 
 # amazon is eventual consistent, retry a bit if failed
 def check_configure_versioning_retry(bucket_name, status, expected_string):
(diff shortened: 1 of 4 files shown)


alimaredia:
@jmunhoz this looks great. Could you add something to the README about this functionality?

jmunhoz:
@alimaredia thanks, added archive zone sync module mention and how to run/enable the related tests.

s3tests_boto3/functional/test_s3.py
```diff
@@ -12133,3 +12138,601 @@ def test_object_lock_uploading_obj():
     eq(response['ObjectLockLegalHoldStatus'], 'ON')
     client.put_object_legal_hold(Bucket=bucket_name, Key=key, LegalHold={'Status':'OFF'})
     client.delete_object(Bucket=bucket_name, Key=key, VersionId=response['VersionId'], BypassGovernanceRetention=True)
+
+
+def _wait_replication_time_az():
+    time.sleep(2)
```
cbodley: i'd really like to find a way to get these tests running in teuthology, but sleeps like this tend to cause a lot of timing failures in the sepia lab environment; performance there is highly variable because we co-locate lots of ceph daemons on the same machine, and run most of them under valgrind

waiting for replication to finish is a problem that we've (mostly) solved in the rgw multisite test suite with the help of radosgw-admin commands. you can find some notes on the test suite in https://github.com/ceph/ceph/blob/master/src/test/rgw/test_multi.md. the zonegroup_meta_checkpoint() and zonegroup_bucket_checkpoint() functions are how we wait for metadata and bucket sync to complete before verifying the results

would you be willing to port these test cases there? @yuvalif did an excellent job adding coverage for the pubsub zone in https://github.com/ceph/ceph/pull/26299, which could serve as an example for the archive zone

@alimaredia raised a separate concern about s3-tests, which aspires to be a generic test suite for all s3 implementations, so we'd like to avoid too many tests that only apply to ceph/rgw

alimaredia: @jmunhoz just out of curiosity how were you orchestrating ceph locally while you were developing these tests?

jmunhoz: @cbodley @alimaredia thanks for reviewing and sharing pointers to test_multi and pubsub patches. I uploaded the new code at https://github.com/ceph/ceph/pull/29359

jmunhoz: > @jmunhoz just out of curiosity how were you orchestrating ceph locally while you were developing these tests?

@alimaredia I set up one only cluster (MON=1 MDS=0 MGR=1 RGW=2 OSD=3) with 1 realm, 1 zonegroup and 2 zones. All of them running on localhost with endpoints on 8001 and 8002 ports. One of the zones is an archive zone.

MON=1 MDS=0 MGR=1 RGW=2 OSD=3

yuvalif: @jmunhoz note that by using the multi_site framework, it would be easy to use the same code for testing setups with more than 2 zones

jmunhoz:
Closing this obsolete PR, it was replaced by https://github.com/ceph/ceph/pull/29359/

Thanks for the review!

PR #291, opened 2019-07-12
Add a test case for continuation token in list objects v2

zhangsw:
This is for https://github.com/ceph/ceph/pull/28988

Signed-off-by: zhang Shaowen <zhangshaowen@cmss.chinamobile.com>

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -1402,11 +1402,14 @@ def test_bucket_listv2_continuationtoken():
     bucket_name = _create_objects(keys=key_names)
     client = get_client()
 
-    response = client.list_objects_v2(Bucket=bucket_name, ContinuationToken='baz')
-    eq(response['ContinuationToken'], 'baz')
-    eq(response['IsTruncated'], False)
-    key_names2 = ['foo', 'quxx']
-    keys = _get_keys(response)
+    response1 = client.list_objects_v2(Bucket=bucket_name, MaxKeys=1)
+    next_continuation_token = response1['NextContinuationToken']
+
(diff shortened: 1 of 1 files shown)


albIN7:
@zhangsw amazon s3 uses startafter to start listing the objects. So, I think it's better to assign startafter as 'baz' and continuationtoken as 'foo' and check whether the remaining key is 'quxx'. Does this makes sense?

zhangsw:
@albIN7 If we specify both startafter and continuationtoken , S3 will ignore the startafter parameter. So as Ceph does. https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html

albIN7:
> @albIN7 If we specify both startafter and continuationtoken , S3 will ignore the startafter parameter. So as Ceph does. https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html

I have tested just now and it is working fine. Could you please test by specifying both startafter and continuationtoken and let me know the output.

albIN7:
@zhangsw I think we should use startafter for listing the first object and continuationtoken for the subsequent objects though continuationtoken works even for listing the first object.

zhangsw:
@albIN7 I add a new test case for specifying both startafter and continuationtoken and it works fine.

alimaredia:
@zhangsw I think we're missing another test where the continuation token is "bad", aka not a string that's in the list of keys. 

@albIN7 what do you think?

zhangsw:
@alimaredia I find that the continuationtoken is obfuscated and is not a real key in S3. If we provide a bad token, S3 will return: ```The continuation token provided is incorrect (Service: Amazon S3; Status Code: 400; Error Code: InvalidArgument)``` At present `continuationtoken` works just like a marker in Ceph. Maybe we can add the case when we make the `continuationtoken` work as S3's.

PR #297, opened 2019-08-06
startafter and continuation token shouldn't  be returned if they are not specified in the request

albin-antony:


albIN7:
@alimaredia @cbodley please review

PR #298, opened 2019-08-07
startafter and continuationtoken are mutually exclusive

albin-antony:


albIN7:
@zhangsw @cbodley @alimaredia Please review

zhangsw:
I think Startafter and ContinuationToken can be specified at same time. If ContinuationToken is provided, Startafter has no effects. I think we need to assure the returned Startafter and ContinuationToken are the same as we specify in the request.

albIN7:
> I think Startafter and ContinuationToken can be specified at same time. If ContinuationToken is provided, Startafter has no effects. I think we need to assure the returned Startafter and ContinuationToken are the same as we specify in the request.

Agree

PR #299, opened 2019-08-23
Check expiration header returns the earliest expiration rule

ofriedma:
Check expiration header returns the earliest expiration rule


s3tests_boto3/functional/test_s3.py
```diff
@@ -9106,6 +9106,104 @@ def test_lifecycle_expiration_header_head():
...
+    bucket_name = get_new_bucket()
+    client = get_client()
+    lifecycle={
+        "Rules": [
+        {
+            "Filter": {
+                "Tag": {"Key": "key1", "Value": "tag1"}
+            },
+            "Status": "Enabled",
+            "Expiration": {
```
mattbenjamin: should we just use the Days element to avoid this breaking somehow in future (or at least looking as if it might)?

ofriedma: It is easier to check if the year 2029 than just adding days. 
It will maybe break in 2029.
Although that it is not using sleep so the object will not be deleted till the head request. 
Can we leave as it is now? 

Thanks

ofriedma: @mattbenjamin I have changed it to on days based.
Is it ok?

Thanks

PR #301, opened 2019-08-27
fix days0 lc expiration test

ofriedma:
The response should be without the key Contents because the bucket should be empty after the lc


mattbenjamin:
@ofriedma , @zhangsw has a PR that prevents LifecycleExpiration from having days 0--and provides evidenced it's correct for AWS.  I added you as a reviewer.

PR #303, opened 2019-09-04
test lifecycle expiration when tags are not the exactly the same as the rule

ofriedma:

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -8900,6 +8900,75 @@ def test_lifecyclev2_expiration():
     eq(len(keep2_objects), 4)
     eq(len(expire3_objects), 2)
 
+@attr(resource='bucket')
+@attr(method='put')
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/test_s3.py
```diff
@@ -8900,6 +8900,75 @@ def test_lifecyclev2_expiration():
     eq(len(keep2_objects), 4)
     eq(len(expire3_objects), 2)
 
+@attr(resource='bucket')
+@attr(method='put')
+@attr(operation='test lifecycle expiration when tags are not the exactly the same as the rule')
```
mattbenjamin: since we'll have this test around forever, we should make the description more clear;  iiuc, the case being tested is one where tag name but not tag value match?

ofriedma: no, when let's say object have 2 tags a:b, d:c and rule is applied on a:b , object should be deleted
The test, test that it will be deleted.

PR #304, opened 2019-09-10
boto2: tag more tests with fails_with_subdomain

cbodley:
test_bucket_policy_different_tenant fails because it adds a : to the beginning of the hostname, ie:
> Host: :test-client.0-oktbnrjh6qacop7-25.s3.smithi003.front.sepia.ceph.com

the append tests fail with 403 Forbidden because they use `key.generate_url()` on a connection with calling_format=subdomain, but send the raw request to the unmodified hostname. because the Host header doesn't match the signature of the presigned url, the signature validation fails

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -155,6 +155,7 @@ def test_versioning_obj_read_not_exist_null():
 @attr(operation='append object')
 @attr(assertion='success')
 @attr('fails_on_aws')
+@attr('fails_with_subdomain')
 @attr('appendobject')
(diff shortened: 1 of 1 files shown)


cbodley:
http://qa-proxy.ceph.com/teuthology/cbodley-2019-09-10_18:28:37-rgw:website-wip-rgw-dns-name-iends-with-distro-basic-smithi/4296622/teuthology.log shows example failures when run with calling_format=subdomain

cbodley:
cherry-picked to ceph-master

PR #306, opened 2019-09-26
Passing sse-kms keys from configuration instead of hard coding in tests

hairesis:
In this change I've tried to trade compatibility between current sse-kms barbican tasks and our feature implementation of https://github.com/ceph/ceph/pull/29783.

Because the way code is  structured* in current sse-kms implementation, s3_tests.py have an handful of duplicated tests which I have removed.
The side effect of it is that, until sse-kms change above is merged, tests for the "testing" sse-kms backend (the local one in the ceph.conf) will not be executed. 

However, all the sse-kms functional tests will run against the Barbican backend. This setup should allow us to run tests for Vault and at the same time not breaking compatibility with current code.

### DIFF
--- a/s3tests_boto3/functional/__init__.py
+++ b/s3tests_boto3/functional/__init__.py
@@ -169,8 +169,11 @@ def setup():
     try:
         config.main_kms_keyid = cfg.get('s3 main',"kms_keyid")
     except (ConfigParser.NoSectionError, ConfigParser.NoOptionError):
-        config.main_kms_keyid = None
-        pass
+        config.main_kms_keyid = 'testkey-1'
+    try:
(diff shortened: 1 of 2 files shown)


alimaredia:
This all looks fine to me. Do you have a teuthology run you can link to show that your changes work?

hairesis:
@alimaredia I've run the tests on a clean master

**Branch for backward compatibility (clean master with tests pointing to this PR)**
https://github.com/hairesis/ceph/blob/wip-s3-tests-backward-compat/qa/suites/rgw/crypt/4-tests/s3tests.yaml

**Teuthology tests**
http://pulpito.front.sepia.ceph.com/h4iresis-2019-10-01_15:54:56-rgw:crypt-master-distro-basic-smithi/

Tests looks green. Can you please review?

PR #267, opened 2019-03-27
Listobjectsv2 testcases

albIN7:
Adding test cases for keycount, delimiter, prefix

PR #307, opened 2019-09-30
Fix validate_bucket_listv2

peter-ginchev:
NextContinuationToken is not cleartext, and we shouldn't assume that.
Also, the test should use it for ContinuationToken


peter-ginchev:
@albin-antony I've fixed #267 which incorrectly assumes NextContinuationToken, please see: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.list_objects_v2

peter-ginchev:
@alimaredia Could you please review this PR, it's a fix to a previous change reviewed by you.

albin-antony:
Thanks @peter-ginchev . LGTM :+1:

PR #308, opened 2019-10-07
Passing sse-kms keys from configuration instead of hard coding in tests

scarvalhojr:
In this change I've tried to trade compatibility between current sse-kms barbican tasks and our feature implementation of Vault support in Rados Gateway (https://github.com/ceph/ceph/pull/29783).

Because the way code is structured* in current sse-kms implementation, s3_tests.py have an handful of duplicated tests which I have removed. The side effect of it is that, until sse-kms change above is merged, tests for the "testing" sse-kms backend (the local one in the ceph.conf) will not be executed.

However, all the sse-kms functional tests will run against the Barbican backend. This setup should allow us to run tests for Vault and at the same time not breaking compatibility with current code.


scarvalhojr:
We'll need to rebase this first

PR #319, opened 2019-11-03
Fixing bootstap for RHEL/Centos 8

### LINKED ISSUE #318: bootstrap is failing in RHEL/Centos 8
bootstrap is failing in RHEL/Centos 8
It looks like the reason is that in  RHEL/Centos 8:
1. `python-virtualenv` was replaced with `python2-virtualenv`
2. `python-devel` was replaced with `python2-devel`
3. `virtualenv` was replaced with `virtualenv-2`

$ ls -la /etc/redhat-release
lrwxrwxrwx 1 root root 14 Aug 14 06:42 /etc/redhat-release -> centos-release

$ cat /etc/redhat-release
CentOS Linux release 8.0.1905 (Core)

$  cd src/test/system_tests/s3-tests/
$ bash -x ./bootstrap
```
...
+ echo './bootstrap: missing required RPM packages. Installing via sudo.'
./bootstrap: missing required RPM packages. Installing via sudo.
+ sudo yum -y install python-virtualenv python-devel libevent-devel libffi-devel libxml2-devel libxslt-devel zlib-devel
CentOS-8 - AppStream                                                                                                                                                                                                                          2.0 MB/s | 6.3 MB     00:03
CentOS-8 - Base                                                                                                                                                                                                                               2.9 MB/s | 7.9 MB     00:02
CentOS-8 - Extras                                                                                                                                                                                                                             893  B/s | 2.1 kB     00:02
No match for argument: python-virtualenv
No match for argument: python-devel
Error: Unable to find a match
```

liranmauda:
Fixing bootstap for RHEL/Centos 8

- When VERSION_ID is less than 8, installing the python packages, otherwise installing the python2 packages.
- When we are using the python2-virtualenv, we will run virtualenv-2.
- Making sure we have which installed

Fixes #318

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -1,8 +1,11 @@
 #!/bin/sh
 set -e
 
+virtualenv="virtualenv"
+packages=(which) 
 if [ -f /etc/debian_version ]; then
-    for package in python-pip python-virtualenv python-dev libevent-dev libffi-dev libxml2-dev libxslt-dev zlib1g-dev; do
+    packages+=(python-pip python-virtualenv python-dev libevent-dev libffi-dev libxml2-dev libxslt-dev zlib1g-dev)
+    for package in ${packages[@]}; do
         if [ "$(dpkg --status -- $package 2>/dev/null|sed -n 's/^Status: //p')" != "install ok installed" ]; then
             # add a space after old values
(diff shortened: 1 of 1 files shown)


liranmauda:
@alimaredia Please review.

PR #322, opened 2019-11-22
Fix check_invalid_bucketname to skip validate_bucket_name entirely

tipabu:


tipabu:
Oh! I _think_ this might also address the

    # TODO: figure out why a 403 is coming out in boto3 but not in boto2.

comment in `test_bucket_create_naming_bad_punctuation`!

PR #278, opened 2019-05-22
Ignore missing x-rgw-* headers

gaul:
Continue to require them in `test_bucket_head_extended`.  Only Ceph
supports these headers.

Signed-off-by: Andrew Gaul <andrew@gaul.org>

PR #323, opened 2019-11-22
Make various tests more likely to pass on AWS

tipabu:
...which of course won't have `x-rgw-object-count` or `x-rgw-bytes-used`.


gaul:
Similar to #278.

tipabu:
https://github.com/ceph/s3-tests/pull/278 merged; good enough for me!

PR #327, opened 2019-11-25
Fix test_bucket_list_prefix_unreadable

tipabu:
Following https://github.com/boto/botocore/pull/726, `list_objects` will automatically include a `encoding-type=url` query param. However, the client does not decode all of the response elements properly -- notably, `Prefix` would remain encoded.

Hopefully this will be fixed soon-ish (I've got a patch proposed at https://github.com/boto/botocore/pull/1901) but in the meantime, use the work-around suggested in https://github.com/boto/boto3/issues/816 of unregistering the `set_list_objects_encoding_type_url` handler.

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -1,4 +1,5 @@
 import boto3
+import botocore.handlers
 import botocore.session
 from botocore.exceptions import ClientError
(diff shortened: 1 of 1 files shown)


tipabu:
Anything else I should do to help move this forward? I'm getting less optimistic about boto fixing the issue...

PR #328, opened 2019-11-26
add test for multipart copy with tenant user

cfanz:
Related to [#42825](https://tracker.ceph.com/issues/42825)
Signed-off-by: cfanz <songxinying@gmail.com>


cfanz:
Although we can request an 'atomic range copy' through rgw rest API, it seems there isn't a corresponding boto3 interface. So one test case for range copy with tenant user in a multipart scenario is enough.

PR #330, opened 2019-12-07
bootstrap: switch to bash

yehudasa:
Use of declare that is bash specific was introduced in:
d44c923405e22ea8b6cc1ab4316be3cc07f29b5


yehudasa:
@cbodley @alimaredia see http://pulpito.ceph.com/yehudasa-2019-12-06_23:57:34-rgw-master-distro-basic-smithi/4579256/

PR #332, opened 2019-12-14
transition to python 3

liewegas:


alimaredia:
This has been fixed in the master and ceph-master branch.

PR #333, opened 2019-12-14
transition to python 3

liewegas:


liewegas:
This doesn't work.. I have no idea what i'm doing

bootstrap
```diff
@@ -34,18 +34,18 @@ elif [ -f /etc/redhat-release ]; then
     fi
 fi
 
-virtualenv --no-site-packages --distribute virtualenv
+virtualenv --no-site-packages --distribute virtualenv -p /usr/bin/python3
```
kshtsk: It is enough to provide `-p python3`.

bootstrap
```diff
@@ -34,18 +34,18 @@ elif [ -f /etc/redhat-release ]; then
...
+#./virtualenv/bin/pip install setuptools==32.3.1
 
 ./virtualenv/bin/pip install -r requirements.txt
 
 # forbid setuptools from using the network because it'll try to use
 # easy_install, and we really wanted pip; next line will fail if pip
 # requirements.txt does not match setup.py requirements -- sucky but
 # good enough for now
-./virtualenv/bin/python setup.py develop
+./virtualenv/bin/python3 setup.py develop
```
kshtsk: there is no need to provide `python3` here, since `./virtualenv/bin/python` is already pointed to python3 created in with `virtualenv`.

s3tests/analysis/rwstats.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
kshtsk: If this code is going to be run in the venv creates as about, there is no sense to use shebang at all.

s3tests/generate_objects.py
```diff
@@ -43,7 +43,7 @@ def upload_objects(bucket, files, seed):
     name_generator = realistic.names(15, 4, seed=seed)
 
     for fp in files:
-        print >> sys.stderr, 'sending file with size %dB' % fp.size
+        sys.stderr.write('sending file with size %dB\n' % fp.size)
```
kshtsk: Here and after for py2/py3 compatibility I guess you better use print from future package, like this:
```
from __future__ import print_function
.
.
.
print('sending file with size %dB' % fp.size, file=sys.stderr)
```

s3tests/generate_objects.py
```diff
@@ -94,18 +94,18 @@ def _main():
 
     bucket.set_acl('public-read')
     keys = []
-    print >> OUTFILE, 'bucket: %s' % bucket.name
-    print >> sys.stderr, 'setup complete, generating files'
+    OUTFILE.write('bucket: %s\n' % bucket.name)
```
kshtsk: See the comment above.

s3tests/realistic.py
```diff
@@ -227,7 +227,7 @@ def files(mean, stddev, seed=None):
     re-download the object and place the contents into a file called
     ``foo``, the following should print two identical lines:
 
-      python -c 'import sys, hashlib; data=sys.stdin.read(); print hashlib.md5(data[:-16]).hexdigest(); print "".join("%02x" % ord(c) for c in data[-16:])' <foo
+      python3 -c 'import sys, hashlib; data=sys.stdin.read(); print hashlib.md5(data[:-16]).hexdigest(); print "".join("%02x" % ord(c) for c in data[-16:])' <foo
```
kshtsk: This code is not python3 compatible.

s3tests_boto3/generate_objects.py
```diff
@@ -43,7 +43,7 @@ def upload_objects(bucket, files, seed):
     name_generator = realistic.names(15, 4, seed=seed)
 
     for fp in files:
-        print >> sys.stderr, 'sending file with size %dB' % fp.size
+        sys.stderr.write('sending file with size %dB\n' % fp.size)
```
kshtsk: See the comment regarding future print.

s3tests_boto3/readwrite.py
```diff
@@ -57,7 +57,7 @@ def reader(bucket, worker_id, file_names, queue, rand):
                         traceback=traceback.format_exc(),
                         ),
                     )
-                print "ERROR:", m
+                print("ERROR:", m)
```
kshtsk: This requires future python feature enabled in order it prints the same way as in py2

s3tests_boto3/realistic.py
```diff
@@ -227,7 +227,7 @@ def files(mean, stddev, seed=None):
     re-download the object and place the contents into a file called
     ``foo``, the following should print two identical lines:
 
-      python -c 'import sys, hashlib; data=sys.stdin.read(); print hashlib.md5(data[:-16]).hexdigest(); print "".join("%02x" % ord(c) for c in data[-16:])' <foo
+      python3 -c 'import sys, hashlib; data=sys.stdin.read(); print hashlib.md5(data[:-16]).hexdigest(); print "".join("%02x" % ord(c) for c in data[-16:])' <foo
```
kshtsk: Looks like dejavu, is it a code duplication here?

tchaikov:
closing. as i think #336 is a better fix.

PR #340, opened 2020-01-15
bootstrap: add support for *suse distributions

theanalyst:
Also reformatted the bootstrap script to parse /etc/os-release instead, so that
more distro/pkg manager support could be added at a later point in time and
fixed the error message

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -3,39 +3,60 @@ set -e
 
 virtualenv="virtualenv"
 declare -a packages
-if [ -f /etc/debian_version ]; then
-    packages=(debianutils python3-pip python3-virtualenv python3-dev libevent-dev libffi-dev libxml2-dev libxslt-dev zlib1g-dev)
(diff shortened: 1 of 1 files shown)


theanalyst:
Also I find some packages like libevent-devel aren't really used at all, not sure where that came from, an old gevent maybe?

alimaredia:
This looks fine and works locally on fedora. Is there a specific package in Suse for python 3.6? The tests don't work on 3.7 or later.

theanalyst:
Nathan mentioned SUSE does not have any way to peg Python 3.6, so I guess we have to stick with this for now until we fix for 3.7 :)

smithfarm:
> Nathan mentioned SUSE does not have any way to peg Python 3.6, so I guess we have to stick with this for now until we fix for 3.7 :)

Well, short of using something like `pythonbrew`, anyway.

PR #341, opened 2020-03-06
Fixed ERROR in test_basic_key_count

brabiega:
Very small fix for the test to FAIL instead of ERROR when expected field is missing from response.

Signed-off-by: Bartosz Rabiega <bartosz.rabiega@corp.ovh.com>

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -190,8 +190,8 @@ def test_basic_key_count():
     for j in range(5):
             client.put_object(Bucket=bucket_name, Key=str(j))
     response1 = client.list_objects_v2(Bucket=bucket_name)
+    eq('KeyCount' in response1, True)
     eq(response1['KeyCount'], 5)
-
 
 
 @attr(resource='bucket')


s3tests_boto3/functional/test_s3.py
```diff
@@ -190,10 +190,10 @@ def test_basic_key_count():
     for j in range(5):
             client.put_object(Bucket=bucket_name, Key=str(j))
     response1 = client.list_objects_v2(Bucket=bucket_name)
+    eq('KeyCount' in response1, True)
     eq(response1['KeyCount'], 5)
```
gaul: If you want to do this, perhaps `response1.get('KeyCount')` avoids repeating the lookup?

brabiega: Double lookup doesn't matter here... It is lightyears faster comparing to other pieces of code. 
I wanted to have better readability without adding more lines of code.

What we want to test here are two things:
1. Some key (KeyCount) is in the response body
2. The value of the key equals 5

And the asserts are directly corresponding to those tests.
At least that's my approach :)

gaul: This is not an efficiency concern, it is a readability and maintenance one.  Look up TOCTOU for other reasons.

PR #342, opened 2020-03-23
commit first tests for s3select and initial framework

galsalomon66:


mattbenjamin:
@galsalomon66 how far is this from proposed merge?

galsalomon66:
> @galsalomon66 how far is this from proposed merge?

not far.
allot is accomplished , but still there are much to do.
we should review the tasks list.

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,411 @@
...
+secret_key = 'b234567890123456789012345678901234567890'
+region_name = ''
+
+def get_connection():
+    conn = boto.connect_s3(
+        aws_access_key_id = access_key,
+        aws_secret_access_key = secret_key,
+        host = 'localhost', port = 8000 ,
+        is_secure=False,               # uncomment if you are not using ssl
+        calling_format = boto.s3.connection.OrdinaryCallingFormat(),
```
cbodley, line 90: all this info needs to come from the [s3tests config file](https://raw.githubusercontent.com/ceph/s3-tests/master/s3tests.conf.SAMPLE). you should probably just use `get_client()` provided by `s3tests_boto3/functional/__init__.py` like test_s3.py does

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+
+    return list_of_int
+       
+def test_count_operation():
+    csv_obj_name = "csv_star_oper"
+    bucket_name = "test"
+    num_of_rows = 10
+    obj_to_load = create_random_csv_object(num_of_rows,10)
+    upload_csv_object(bucket_name,csv_obj_name,obj_to_load)
+    res = remove_xml_tags_from_result( run_s3select(bucket_name,csv_obj_name,"select count(0) from stdin;") ).replace(",","")
```
yuvalif, line 181: is it expected by the spec to have the extra comma at the end of the result set?

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+    
+    result = ""
+    for event in r['Payload']:
+        if 'Records' in event:
+            records = event['Records']['Payload'].decode('utf-8')
+            result += records
+    
+    return result
+
+def remove_xml_tags_from_result(obj):
```
yuvalif, line 152: why not run the result through a standard xml parser?
this would also validate that the xml format is correct, and what would any client implementation do

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+        result = ""
+        for i in range(rows):
+            row = "";
+            for y in range(columns):
+                row = row + "{}{:02d}{:02d}-{:02d}{:02d}{:02d},".format(random.randint(0,100)+1900,random.randint(1,12),random.randint(1,28),random.randint(0,23),random.randint(0,59),random.randint(0,59),);
+            result += row + "\n"
+
+        return result
+
+def create_random_csv_object(rows,columns,col_delim=",",record_delim="\n",csv_schema=""):
```
yuvalif, line 103: what about string columns?

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+            for y in range(columns):
+                row = row + "{}{}".format(random.randint(0,1000),col_delim);
+            result += row + record_delim
+
+        return result
+
+
+def upload_csv_object(bucket_name,new_key,obj):
+        conn = get_connection()
+        conn.create_bucket( bucket_name )
```
yuvalif, line 119: why do you create the bucket inside the upload function?
we should test multiple objects in the same bucket as well

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+
+    csv_obj_name = "csv_10000x10"
+    bucket_name = "test"
+    upload_csv_object(bucket_name,csv_obj_name,csv_obj)
+    
+    res_s3select = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,"select min(int(_1)) from stdin;")  ).replace(",","")
+    list_int = create_list_of_int( 1 , csv_obj )
+    res_target = min( list_int )
+
+    assert int(res_s3select) == int(res_target)
```
yuvalif, line 196: use assert_equals from nosetest

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+    csv_obj = create_random_csv_object(10000,10)
+
+    csv_obj_name = "csv_100000x10"
+    bucket_name = "test"
+    upload_csv_object(bucket_name,csv_obj_name,csv_obj)
+
+    res_s3select = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,"select min(int(_1)),max(int(_2)),min(int(_3))+1 from stdin;")).replace("\n","")
+
+    # assert is according to radom-csv function 
+    assert res_s3select == "0,1000,1,"
```
yuvalif, line 240: there is some probability that this would fail - please avoid in automated tests

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+    
+    csv_obj = create_random_csv_object(10000,10)
+
+    csv_obj_name = "csv_10000x10"
+    bucket_name = "test"
+    upload_csv_object(bucket_name,csv_obj_name,csv_obj)
+
+    res_s3select_alias = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,"select int(_1) as a1,int(_2) as a2, a1+a4 as a3, a5+a1 as a4, int(_3)+a3 as a5 from stdin;")  )
+
+    find_res = res_s3select_alias.find("number of calls exceed maximum size, probably a cyclic reference to alias");
```
yuvalif, line 285: if the error text would change we would have an issue here. probably better to just assert on the action failure, or code

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+    
+    assert int(find_res) >= 0 
+
+def test_datetime():
+
+    # purpose of test is to validate date-time functionality is correct,
```
yuvalif, line 291: random tests are great! but, in addition, we can add more accurate tests, where the input csv is not randomly generated, and the result set is checked against the expected result

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+
+    csv_obj = ',first,,,second,third="c31,c32,c33",forth="1,2,3,4",fifth="my_string=\\"any_value\\" , my_other_string=\\"aaaa,bbb\\" ",' + "\n"
+    csv_obj_name = "csv_one_line"
+    bucket_name = "test"
+
+    upload_csv_object(bucket_name,csv_obj_name,csv_obj)
+
+    # return value contain comma{,}
+    res_s3select_alias = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,"select _6 from stdin;")  ).replace("\n","")
+    assert res_s3select_alias ==  'third="c31,c32,c33",'
```
yuvalif, line 335: when you have unescaped quotes, don't you remove the quotes?

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+    assert res_s3select_date_time == res_s3select_count
+
+    # validate that utcnow is integrate correctly with other date-time functions 
+    res_s3select_date_time_utcnow = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select count(0) from  stdin where datediff("hours",utcnow(),dateadd("day",1,utcnow())) == 24 ;')  )
+
+    assert res_s3select_date_time_utcnow == res_s3select_count
+
+def test_csv_parser():
+
+    # purpuse: test default csv values(, \n " \ ), return value may contain meta-char
```
yuvalif, line 324: could you please also verify:
* newline inside a quoted columns
* escaped escape charecter
could look here for a [list of ideas](https://github.com/maxogden/csv-spectrum/tree/master/csvs)

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,410 @@
...
+    csv_obj_name = "csv_with_header_info"
+    bucket_name = "test"
+
+    upload_csv_object(bucket_name,csv_obj_name,csv_obj)
+
+    # ignoring the schema on first line and retrieve using generic column number
+    res_ignore = remove_xml_tags_from_result( run_s3select(bucket_name,csv_obj_name,"select _1,_3 from stdin;",csv_header_info="IGNORE") ).replace("\n","")
+
+    # using the scheme on first line, query is using the attach schema
+    res_use = remove_xml_tags_from_result( run_s3select(bucket_name,csv_obj_name,"select c1,c3 from stdin;",csv_header_info="USE") ).replace("\n","")
```
yuvalif, line 396: probably good to create a separate schema tests, including use of aliases

yuvalif:
please add some more negative tests:
* syntax errors
* type errors (e.g. arithmetic operations on strings)
* CSV formatting issues (different columns in different rows)
* substr out of bound (negative index, large index)
* invalid date format
* CSV schema mismatch

s3tests_boto3/functional/test_s3select.py
galsalomon66: output serialization is not implemented (only input serialization is implemented, part of it).
the comma at the end, is not part of spec (row delimiter , field delimiter should be parameterized)

s3tests_boto3/functional/test_s3select.py
galsalomon66: upon output serialization is implemented, maybe it's better to run standard xml-parser.
currently the /s3.select_object_content/ validates correct output.

s3tests_boto3/functional/test_s3select.py
galsalomon66: this function create outputs, for testing the integration between {rows X columns} and s3-select query processing.
there are tests for checking CSV different use-cases.

s3tests_boto3/functional/test_s3select.py
galsalomon66: it's actually uploading different object to the same bucket of different tests (there is no delete-objects).

s3tests_boto3/functional/test_s3select.py
galsalomon66: true, but currently error-coding is not implemented.
currently, test must validate that correct error-flow is executed.

upon upgrading/changing error-flow into more standard procedure (error-code, error-message-file), test will modified.

s3tests_boto3/functional/test_s3select.py
galsalomon66: generally many tests are missing, there is no full coverage.

s3-select functionalities is quite complex, each function may contain several flows(including error-flows)
thus,accurate test with accurate result, should added.

please note, that with random input (datetime in this case) , combine with complex query (datediff,dateadd,date-extract,substr), actually validate result correctness.

s3tests_boto3/functional/test_s3select.py
galsalomon66: it's in my list to check, not completely sure about that.

s3tests_boto3/functional/test_s3select.py
galsalomon66: escape-escaped is not implemented (i.e. successive escape characters ) 

newline (row-delimiter?) within quotes , if you meant "0xa" , in quotes , it's not supported (there is a test to validate that), since it may reduce performance (search for closing quote).

as for control-chars with escape (specific char sequence) , such as \t,\r,\n its not implemented.

"list of ideas" is ok, but need to priorities implementation efforts.

galsalomon66:
> please add some more negative tests:
> 
> * syntax errors
> * type errors (e.g. arithmetic operations on strings)
> * CSV formatting issues (different columns in different rows)
> * substr out of bound (negative index, large index)
> * invalid date format
> * CSV schema mismatch

s3tests_boto3/functional/test_s3select.py
yuvalif: if the comma at the end is not expected, can we remove it?

s3tests_boto3/functional/test_s3select.py
yuvalif: where should "output serialization" be implemented?
is this in the client or in the product?
my suggestion here is to run the output via an xml parser, to make sure it is correct

s3tests_boto3/functional/test_s3select.py
yuvalif: the random generator, create only numeric columns, can we have tests against csv with string columns as well. this is unrelated to csv use-cases, this would verify string functions, operators, where clause on string columns etc.

s3tests_boto3/functional/test_s3select.py
yuvalif: ok, juts make sure that the test covers multiple objects in multiple buckets

s3tests_boto3/functional/test_s3select.py
yuvalif: if we don't have error codes, just validate that we got an error message. these kinds of assertions tend to cause flaky tests in the future

s3tests_boto3/functional/test_s3select.py
yuvalif: did you look here: https://en.wikipedia.org/wiki/Comma-separated_values
its pretty comprehensive

s3tests_boto3/functional/test_s3select.py
yuvalif: we don't have to implement all aspect of CSV, but we must document exactly what we implemented. 
if we didn't implement the whole feature, our documentation cannot just add a reference to a spec (from the RFC, or AWS)

s3tests_boto3/functional/test_s3select.py
galsalomon66: ok

s3tests_boto3/functional/test_s3select.py
galsalomon66: "output serialization" implemented for the AWS cli (here its boto3, same).
its part of spec implemented in RGW (header-type,message-type,crc,and xml tags).

s3tests_boto3/functional/test_s3select.py
galsalomon66: by default all tokens are string (for CSV), in order to get it as number it must be cast first.
indeed the current framework generate only numbers, but there are actually strings.
there are tests including where clause with substr functions.

```python
    res_s3select_substr = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select min(int(_2)),max(int(_2)) from stdin where substr(_2,1,1) == "1"')).replace("\n","")
```

s3tests_boto3/functional/test_s3select.py
galsalomon66: ok

s3tests_boto3/functional/test_s3select.py
galsalomon66: agreed, once someone will modify the alias functionality and error flow, it may fail.
but, must run s3-tests following any change, before commiting.
the change of error-code/message quite comprehensive, it also change the output serialization.

s3tests_boto3/functional/test_s3select.py
galsalomon66: in documentation there is a section(a table) describing different CSV parsing scenario's.
in case something is missing, i will add there.

s3tests_boto3/functional/test_s3select.py
galsalomon66: we may follow that, or RFC, its product  decision. (not clear what rules AWS followed)
current ceph-s3-select documentation is "copied" from CSV test scenario.
its not "locked", the CSV transition table is flexible and may change.

s3tests_boto3/functional/test_s3select.py
galsalomon66: as mentioned, there is detailed table (in dedicated section) for CSV different cases.(more detailed than AWS)
ceph-s3-select is currently not compliant with AWS, we will probably need to change that once we integrate with data-analytic utility such as presto(for one example).

s3tests_boto3/functional/test_s3select.py
yuvalif: > by default all tokens are string (for CSV), in order to get it as number it must be cast first.
> indeed the current framework generate only numbers, but there are actually strings.
> there are tests including where clause with substr functions.
> 
> ```python
>     res_s3select_substr = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select min(int(_2)),max(int(_2)) from stdin where substr(_2,1,1) == "1"')).replace("\n","")
> ```

what about string operators (e.g. str+str)?

s3tests_boto3/functional/test_s3select.py
yuvalif: there is no issue with the current scope of CSV implementation, especially as it  pretty comprehensive, and would be quite easy to extend it in the future.
but, we must test and document whatever set of feature we currently support

s3tests_boto3/functional/test_s3select.py
galsalomon66: + is not implemented for string.

s3tests_boto3/functional/test_s3select.py
yuvalif: ok. looked at the table again, i think it coverts everything (including our newline in quote behavior)

s3tests_boto3/functional/test_s3select.py
yuvalif: there is no need to validate the output ourselves if the boto3 client already does that
e.g. if we reply invalid xml tags or structure the boto3 client would error?

galsalomon66: yes.
in the course of output serialization implementation, it reject RGW return response (after accepting request and processing it), unless it comply to client(AWS) specification.

did not try to validate response with xml-parser.
the other important clients would be analytical utilities such as presto.

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,525 @@
...
+        
+    
+def run_s3select(bucket,key,query,column_delim=",",row_delim="\n",quot_char='"',esc_char='\\',csv_header_info="NONE"):
+
+    s3 = boto3.client('s3',#'sns',
+        endpoint_url=get_config_endpoint(),
+        aws_access_key_id=get_main_aws_access_key(),
+        region_name=region_name,
+        aws_secret_access_key=get_main_aws_secret_key())
+        #config=Config(signature_version='v2'))
```
cbodley, line 217: please just use `get_client()` instead

s3tests_boto3/functional/test_s3select.py
```diff
@@ -0,0 +1,525 @@
...
+
+from .utils import assert_raises
+from .utils import generate_random
+from .utils import _get_status_and_error_code
+from .utils import _get_status
+
+from .policy import Policy, Statement, make_json_policy
+
+from . import (
+    get_client,
```
cbodley, line 39: tons of unused imports here. get_client() is probably the only one you need?

cbodley:
cherry-picked these commits to ceph-master

PR #302, opened 2019-08-30
Public buckets

theanalyst:
tests for https://github.com/ceph/ceph/pull/30033

PR #343, opened 2020-03-26
Ceph master public buckets backport

theanalyst:
ceph-master backport of #302


cbodley:
@theanalyst still passing against master? if so, please merge!

theanalyst:
Yes the last tests were done against ceph-master with these commits against master

PR #345, opened 2020-05-05
Fix: Bucket resource leak when cleanup

huangp0600:
In the function of nuke_prefixed_buckets, if some err is thrown when deleting buckets, the left buckets remain uncleaned.
It is kind of resoruce leak on some charged platform. We have to clear them manually.

I know the original code is meant to give the user some hint by rasing error. But the resource leak of left buckets is a little annoying.

Signed-off-by: Pei <huangp0600@126.com>

### DIFF
--- a/s3tests_boto3/functional/__init__.py
+++ b/s3tests_boto3/functional/__init__.py
@@ -113,6 +113,7 @@ def nuke_prefixed_buckets(prefix, client=None):
 
     buckets = get_buckets_list(client, prefix)
 
+    err = None
     if buckets != []:
         for bucket_name in buckets:
             objects_list = get_objects_list(bucket_name, client)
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/__init__.py
```diff
@@ -127,9 +127,6 @@ def nuke_prefixed_buckets(prefix, client=None):
             try:
                 response = client.delete_bucket(Bucket=bucket_name)
             except ClientError:
-                # if DELETE times out, the retry may see NoSuchBucket
-                if response['Error']['Code'] != 'NoSuchBucket':
-                    raise ClientError
```
gaul, line 132: What if you save the first or last client error and re-raise it after the loop completes?

huangp0600: For example, 
```
error A occurs when delete bucket-1
...
error B occurs when delete bucket-2
```
If we save the last client error and re-raise it after the loop completes, user knows there is an exception B. then he debugs and make some effort to resolve it.
then he have to run s3tests again to know about exception A. 
Maybe this path is a little time-cost or inefficiency.

Can we print every err like this:
```
try:
    response = client.delete_bucket(Bucket=bucket_name)
except ClientError:
    print response
    pass
```

gaul: If the error is not raised then no one will look at the logs...

huangp0600: > If the error is not raised then no one will look at the logs...

Agree. I would save the last client error and re-raise it after the loop completes.

zhiteng:
@cbodley @alimaredia Could you please take a look this fix and help merge this?  Thanks.

alimaredia:
@huangp0600 this looks fine to me. Could you just put some comments in the new code explaining the reason for the change or the new code briefly

huangp0600:
> @huangp0600 this looks fine to me. Could you just put some comments in the new code explaining the reason for the change or the new code briefly
@alimaredia , thanks. Explanation has been added for the change as a comment

PR #347, opened 2020-05-19
fix/restore test_lifecycle_expiration checks

mattbenjamin:
Commit bf956df71e56bcdae7d41b9789d8e8d775f76007 adding
listobvjectsv2 tests inadvertently changed the v1
test_lifecycle_expiration test, which it had copied to
create a v2 version.  Revert this.


cbodley:
it seems silly to have different versions of lifecycle expiration tests for list v1/v2; the bucket listing is just to verify the contents, right?

cbodley:
cherry-picked to ceph-master

PR #352, opened 2020-07-03
per each new uploaded file(for test), it got unique name(random), and…

galsalomon66:
… uploaded file is verified for its content

Signed-off-by: gal salomon <gal.salomon@gmail.com>


galsalomon66:
the main modification for tests, is unique name per object.

PR #354, opened 2020-07-18
Use marker when calling list_objects

gaul:
Previously `test_bucket_list_delimiter_not_skip_special` failed in
`nuke_buckets` due to not deleting all the keys.


gaul:
@cbodley Please review

PR #355, opened 2020-07-30
STS Tests Files and modification in __init__.py

TRYTOBE8TME:
Signed-off-by: Kalpesh Pandya <kapandya@redhat.com>

### DIFF
--- a/s3tests_boto3/functional/__init__.py
+++ b/s3tests_boto3/functional/__init__.py
@@ -158,6 +158,8 @@ def setup():
         raise RuntimeError('Your config file is missing the "s3 alt" section!')
     if not cfg.has_section("s3 tenant"):
         raise RuntimeError('Your config file is missing the "s3 tenant" section!')
+    if not cfg.has_section("iam"):
+        raise RuntimeError('Your config file is missing the "iam" section!')
 
     global prefix
 
@@ -205,6 +207,12 @@ def setup():
     config.tenant_user_id = cfg.get('s3 tenant',"user_id")
     config.tenant_email = cfg.get('s3 tenant',"email")
 
+    config.iam_access_key = cfg.get('iam',"access_key")
+    config.iam_secret_key = cfg.get('iam',"secret_key")
+    config.iam_display_name = cfg.get('iam',"display_name")
+    config.iam_user_id = cfg.get('iam',"user_id")
+    #config.iam_email = cfg.get('iam',"email")
+
     # vars from the fixtures section
     try:
         template = cfg.get('fixtures', "bucket prefix")
@@ -244,6 +252,32 @@ def get_v2_client():
                         endpoint_url=config.default_endpoint,
                         use_ssl=config.default_is_secure,
                         config=Config(signature_version='s3'))
+    return client
+
+def get_sts_client(client_config=None):
+    if client_config == None:
+        client_config = Config(signature_version='s3v4')
+
+    client = boto3.client(service_name='sts',
+                        aws_access_key_id=config.alt_access_key,
+                        aws_secret_access_key=config.alt_secret_key,
+                        endpoint_url=config.default_endpoint,
+                        region_name='',
+                        use_ssl=config.default_is_secure,
+                        config=client_config)
+    return client
+
+def get_iam_client(client_config=None):
+    if client_config == None:
+        client_config = Config(signature_version='s3v4')
+    
+    client = boto3.client(service_name='iam',
+                        aws_access_key_id=config.iam_access_key,
+                        aws_secret_access_key=config.iam_secret_key,
+                        endpoint_url=config.default_endpoint,
+                        region_name='',
+                        use_ssl=config.default_is_secure,
+                        config=client_config)
     return client
 
 def get_alt_client(client_config=None):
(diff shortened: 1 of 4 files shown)


s3tests_boto3/functional/test_sts.py
```diff
@@ -0,0 +1,253 @@
...
+
+@attr(resource='get session token')
+@attr(method='get')
+@attr(operation='check')
+@attr(assertion='s3 ops only accessible by temporary credentials')
+def test_get_session_token():
+    response=""
+    resp=""
+    resp_err=None
+    s3bucket=""
```
alimaredia, line 87: @TRYTOBE8TME I don't think these variables need to be initialized for every test, I don't think the s3-tests are doing so

TRYTOBE8TME: Okay, I'll change this.

s3tests_boto3/functional/test_sts.py
```diff
@@ -0,0 +1,228 @@
...
+        role_response = iam_client.put_user_policy(UserName=username,PolicyName=policyname,PolicyDocument=policy_document)
+    except ClientError as e:
+        role_err = e.response['Code']
+    return (role_err,role_response)
+
+@attr(resource='get session token')
+@attr(method='get')
+@attr(operation='check')
+@attr(assertion='s3 ops only accessible by temporary credentials')
+def test_get_session_token():
```
alimaredia, line 80: For all of your tests can you add a filter similar to this one: https://github.com/ceph/s3-tests/blob/master/s3tests/functional/test_s3.py#L223 to make sure that your tests don't get run with all of the s3-tests and that your tests only get run with the specified filter. Could you also add a section to the README on these sts tests?

TRYTOBE8TME: Yes sure, I'll add the sections and do the required changes in the README.

s3tests_boto3/functional/__init__.py
```diff
@@ -205,6 +207,12 @@ def setup():
     config.tenant_user_id = cfg.get('s3 tenant',"user_id")
     config.tenant_email = cfg.get('s3 tenant',"email")
 
+    config.iam_access_key = cfg.get('iam',"access_key")
+    config.iam_secret_key = cfg.get('iam',"secret_key")
+    config.iam_display_name = cfg.get('iam',"display_name")
+    config.iam_user_id = cfg.get('iam',"user_id")
+    #config.iam_email = cfg.get('iam',"email")
```
alimaredia, line 214: if this line is commented out, can you remove it?

TRYTOBE8TME: I'm not using this but if someone wants then they can use it that's the reason why I thought keeping it.

s3tests_boto3/functional/test_sts.py
```diff
@@ -0,0 +1,233 @@
...
+    sts_user_id=get_alt_user_id()
+    default_endpoint=get_config_endpoint()
+    role_session_name=get_parameter_name()
+    policy_document = "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Principal\":{\"AWS\":[\"arn:aws:iam:::user/"+sts_user_id+"\"]},\"Action\":[\"sts:AssumeRole\"]}]}"
+    (role_error,role_response,general_role_name)=create_role(iam_client,'/',None,policy_document,None,None,None)
+    eq(role_response['Role']['Arn'],'arn:aws:iam:::role/'+general_role_name+'')
+    role_policy = "{\"Version\":\"2012-10-17\",\"Statement\":{\"Effect\":\"Allow\",\"Action\":\"s3:*\",\"Resource\":\"arn:aws:s3:::*\"}}"
+    (role_err,response)=put_role_policy(iam_client,general_role_name,None,role_policy)
+    eq(response['ResponseMetadata']['HTTPStatusCode'],200)
+    resp=sts_client.assume_role(RoleArn=role_response['Role']['Arn'],RoleSessionName=role_session_name,DurationSeconds=900)
```
alimaredia, line 218: @TRYTOBE8TME @pritha-srivastava why is the time set to 900 seconds (15 minutes)? Could we make that number lower somehow?

TRYTOBE8TME: Actually this test checks that the temporary credentials expire after the said time. And 900 secs is the minimum time required. Thus we need to keep this.

alimaredia: Just out of curiosity where is that 900 second minimum set?

s3tests_boto3/functional/__init__.py
alimaredia: If there's no use for it right now in any of the tests could you remove it?

s3tests_boto3/functional/__init__.py
```diff
@@ -246,6 +254,32 @@ def get_v2_client():
...
+
+def get_iam_client(client_config=None):
+    if client_config == None:
+        client_config = Config(signature_version='s3v4')
+    
+    client = boto3.client(service_name='iam',
+                        aws_access_key_id=config.iam_access_key,
+                        aws_secret_access_key=config.iam_secret_key,
+                        endpoint_url=config.default_endpoint,
+                        region_name='',
```
alimaredia, line 278: Why is `region_name` set here? It's not set in any of the other get_client functions

s3tests_boto3/functional/test_sts.py
TRYTOBE8TME: In the following link: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sts.html#STS.Client.assume_role you can see that in Duration Seconds section.

TRYTOBE8TME:
Okay, I'll think over this after a while because in some days I might also raise a PR for the 3rd API tests. So, I guess after this we can properly generalize them into some functions and have these functions in sts_utils.py. So, would this be fine?

PR #364, opened 2020-10-13
s3select tests for coalesce and case

albin-antony:

### DIFF
--- a/s3tests_boto3/functional/test_s3select.py
+++ b/s3tests_boto3/functional/test_s3select.py
@@ -450,6 +450,82 @@ def test_schema_definition():
     assert res_multiple_defintion.find("multiple definition of column {c4} as schema-column and alias") > 0
 
 @attr('s3select')
+def test_when_than_else_expressions():
+
+    csv_obj = create_random_csv_object(10000,10)
+
+    csv_obj_name = get_random_string()
+    bucket_name = "test"
+    upload_csv_object(bucket_name,csv_obj_name,csv_obj)
+
+    res_s3select = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select  case when ((4*3)==(12)) than "case_1_2" else "case_2_1" end from stdin where (3*3==9);')  ).replace("\n","")          
+
+    nose.tools.assert_equal( res_s3select, "case_1_2,")
+
+    res_s3select = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select case when cast(_1 as int)>100 and cast(_1 as int)<200 than "(100-200)" when cast(_1 as int)>200 and cast(_1 as int)<300 than "(200-300)" else "NONE" end from s3object;')  ).replace("\n","")
+
+    count1 = res_s3select.count("(100-200)")  
+
+    count2 = res_s3select.count("(200-300)") 
+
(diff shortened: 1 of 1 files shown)


galsalomon66:
using constant values in query is less desire in s3-tests  (its done on unit tests)
s3-tests is not just testing the aws-cli, it tests queries on many-rows-columns (random values).

s3tests_boto3/functional/test_s3select.py
```diff
@@ -449,6 +449,44 @@ def test_schema_definition():
...
+
+    nose.tools.assert_equal( res_s3select, "5,")
+
+    res_s3select = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select coalesce(nullif(5,5),nullif(1,1.0),2) from stdin;')  ).replace("\n","")   
+
+    nose.tools.assert_equal( res_s3select, "2,")
+
+    res_s3select = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select count(*) from stdin where  coalesce(nullif(7.0,7),nullif(4,4.0),6) == 6;')  ).replace("\n","")   
+
+    nose.tools.assert_equal( res_s3select, "1,")
```
galsalomon66, line 488: ok, 
beside the constant values, we had discussed.
the **coalesce** should test for **order of processing**, meaning,
in case it got more than one non-null values , is it return the correct one?

albin-antony: Yes, I have tried some unit tests. It always returns the first non-null value

galsalomon66:
please modify those tests.
tests should read , process and produce result according to content of objects.

galsalomon66:
for the complex statement /case-when-then/ there is no multi row test. 
only single row test.

for one example, its possible to test as follows

select case when cast(_1 as int)>100 and cast(_1 as int)<200 than "(100-200)" when cast(_1 as int)>200 and cast(_1 as int)<300 than "(200-300)" else "NONE" end from s3object

the query will produce 3 different print-out , it's possible to count result with **_string.count( substring )_**

and compare against select count(0) where ... per each when-condition

s3tests_boto3/functional/test_s3select.py
```diff
@@ -449,6 +449,82 @@ def test_schema_definition():
...
+    bucket_name = "test"
+    upload_csv_object(bucket_name,csv_obj_name,csv_obj)
+
+    res_s3select = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select  case when ((4*3)==(12)) than "case_1_2" else "case_2_1" end from stdin where (3*3==9);')  ).replace("\n","")          
+
+    nose.tools.assert_equal( res_s3select, "case_1_2,")
+
+    res_s3select = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select case when cast(_1 as int)>100 and cast(_1 as int)<200 than "(100-200)" when cast(_1 as int)>200 and cast(_1 as int)<300 than "(200-300)" else "NONE" end from s3object;')  ).replace("\n","")
+
+    count1 = res_s3select.count("(100-200)")
```
galsalomon66, line 467: ok

galsalomon66 (approved):

PR #365, opened 2020-11-03
Tests for multi-object delete with object lock

markhoughton-microfocus:
Tests to go along with https://github.com/ceph/ceph/pull/37933, which fixes an issue where multi-object delete didn't respect object locks.


cbodley (approved):

cbodley:
squashed and rebased, merged, and cherry-picked to ceph-master

PR #373, opened 2020-12-05
rename charlength and character_length function names

galsalomon66:
Signed-off-by: galsalomon66 <gal.salomon@gmail.com>

### DIFF
--- a/s3tests_boto3/functional/test_s3select.py
+++ b/s3tests_boto3/functional/test_s3select.py
@@ -344,7 +344,7 @@ def test_in_expressions():
 
     nose.tools.assert_equal( res_s3select_in, res_s3select )
 
-    res_s3select_in = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select int(_1) from stdin where characterlength(_1) == 2 and substr(_1,2,1) in ("3");')).replace("\n","")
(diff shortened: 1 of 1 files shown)


galsalomon66:
some of the tests which uses string-length functions failed since the s3select-engine was updated (to be align with AWS function names)
the like-tests uses much bigger object (which is not that big ..), but in some of teuthology runs it may cause timeout on boto3.
to avoid those random failures in teuthology, the object size was reduced.

yuvalif:
can you please put a link to the PR where the code change is done?

s3tests_boto3/functional/test_s3select.py
```diff
@@ -612,10 +612,6 @@ def test_when_than_else_expressions():
     bucket_name = "test"
     upload_csv_object(bucket_name,csv_obj_name,csv_obj)
 
-    res_s3select = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select  case when ((4*3)==(12)) than "case_1_2" else "case_2_1" end from stdin where (3*3==9);')  ).replace("\n","")
```
yuvalif, line 615: why is this test removed?

galsalomon66: the test is wrong , the assert for that test is wrong , it should compare against result of other query.
(in later PR https://github.com/ceph/s3-tests/pull/375 , it is fixed)

galsalomon66:
> can you please put a link to the PR where the code change is done?

https://github.com/ceph/s3select/pull/31/files#diff-df8fff244ae0355c2313b8bf7c2f6e353d791f6ac939620a4e9c80b877519580R88

yuvalif (approved):

PR #381, opened 2021-01-13
change test_s3select.py permission; add s3select attribute

galsalomon66:
Signed-off-by: galsalomon66 <gal.salomon@gmail.com>


galsalomon66:
file permission avoid from test_s3select.py to run

galsalomon66 (commented):
minor and crucial change
file permission (cause to skip the test file)
s3select function name (cause test failure due to syntax error)

alimaredia (approved):

PR #387, opened 2021-03-11
add test for dash/underscore handling in keys

mattbenjamin:
The dash is implied in many tests, but included for clarity


s3tests_boto3/functional/test_s3.py
```diff
@@ -2152,6 +2152,17 @@ def test_object_set_get_metadata_overwrite_to_empty():
     got = _set_get_metadata('', bucket_name)
     eq(got, '')
 
+@attr(resource='object.metadata')
+@attr(method='put')
+@attr(operation='metadata key with underscore')
+@attr(assertion='correct handling of dash and underscore in key names')
+def test_object_set_get_metadata_dash_underscore():
+    bucket_name = get_new_bucket()
+    got = _set_get_metadata('kay-fabe')
```
cbodley, line 2161: since you're creating a bucket for these, you should pass `bucket_name` in the calls to `_set_get_metadata(metadata, bucket_name=None)`

mattbenjamin: yes, of course--I was mixing them

cbodley (approved):

PR #389, opened 2021-04-07
sts: test role policy around nonexistent objects

cbodley:
two test cases for HEAD on a nonexistent object: one where the assumed role policy allows s3:ListBucket, and where it denies


adamemerson (approved):
Looks Good To Me.

cbodley:
ok, both tests are passing against https://github.com/ceph/ceph/pull/40631. when run against master, `test_assume_role_allow_head_nonexistent` fails as expected

PR #390, opened 2021-04-25
test extra complete_multipart_upload()

mkogan1:
after the 1st successfull one should also return 200 OK

related tracker issue: https://tracker.ceph.com/issues/50141
related pr: https://github.com/ceph/ceph/pull/40594

Signed-off-by: Mark Kogan <mkogan@redhat.com>


cbodley (approved):

cbodley:
let's give this a couple days before merging, so we don't introduce test failures for other PRs that haven't rebased on https://github.com/ceph/ceph/pull/40594 yet

PR #393, opened 2021-05-13
test multipart upload with bucket policy

ofriedma:
test multipart upload with bucket policy


s3tests_boto3/functional/test_s3.py
```diff
@@ -13135,3 +13135,34 @@ def test_ignore_public_acls():
...
+    policy_document = json.dumps(
+    {
+        "Version": "2012-10-17",
+        "Statement": [{
+        "Effect": "Allow",
+        "Principal": "*",
+        "Action": "*",
+        "Resource": [
+            "{}".format(resource1),
+            "{}".format(resource2)
```
cbodley, line 13159: nit: why is format needed here, can't we just use `[resource1, resource2]`?

cbodley (approved):

s3tests_boto3/functional/test_s3.py
ofriedma: Fixed

PR #394, opened 2021-05-25
s3select: align s3select tests with ceph

galsalomon66:
Update s3-tests to handle the error-response (return 400, and error-description)


galsalomon66:
#402 is handling all relevant tests

PR #397, opened 2021-07-06
sts: Addition of new STS tests

TRYTOBE8TME:
Signed-off-by: Kalpesh Pandya <kapandya@redhat.com>

PR #398, opened 2021-07-08
[DNM]rgw/sts: adding sts tests related to session policy

pritha-srivastava:
using role policy, session policy and bucket policy
for ops like putobj, getobj and copyobj.

Signed-off-by: Pritha Srivastava <prsrivas@redhat.com>

### DIFF
--- a/s3tests_boto3/functional/__init__.py
+++ b/s3tests_boto3/functional/__init__.py
@@ -374,6 +374,36 @@ def get_svc_client(client_config=None, svc='s3'):
 
 bucket_counter = itertools.count(1)
 
+def get_s3_client_using_iam_creds(client_config=None):
+    cfg = configparser.RawConfigParser()
+    try:
+        path = os.environ['S3TEST_CONF']
+    except KeyError:
+        raise RuntimeError(
+            'To run tests, point environment '
+            + 'variable S3TEST_CONF to a config file.',
+            )
+    cfg.read(path)
+    if not cfg.has_section("iam"):
+        raise RuntimeError('Your config file is missing the "iam" section!')
+
+    config.iam_access_key = cfg.get('iam',"access_key")
+    config.iam_secret_key = cfg.get('iam',"secret_key")
+    config.iam_display_name = cfg.get('iam',"display_name")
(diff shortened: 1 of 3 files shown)


s3tests_boto3/functional/__init__.py
```diff
@@ -374,6 +374,36 @@ def get_svc_client(client_config=None, svc='s3'):
 
 bucket_counter = itertools.count(1)
 
+def get_s3_client_using_iam_creds(client_config=None):
```
TRYTOBE8TME, line 377: I think it would be good if we use this function in our test_sts.py file?

s3tests_boto3/functional/test_sts.py
```diff
@@ -354,3 +386,361 @@ def test_assume_role_with_web_identity_invalid_webtoken():
...
+    OpenIDConnectProviderArn=oidc_arn
+    )
+
+@attr(resource='assume role with web identity')
+@attr(method='put')
+@attr(operation='check')
+@attr(assertion='checking session policy working')
+@attr('webidentity_test')
+@attr('session_policy')
+def test_session_policy_bucket_access_not_permitted():
```
TRYTOBE8TME, line 552: It seems similar to https://github.com/ceph/s3-tests/pull/397/files#diff-ab111785021219d559b4dc9596ca595d25243d21624e1c53ffaf5bee8dd6b287R382

So, should I remove this test from my PR?

s3tests_boto3/functional/__init__.py
pritha-srivastava: I have used it, but I like your version better, once all other comment are addressed, I will rebase my commits on top of yours use the version that you have defined.

s3tests_boto3/functional/test_sts.py
pritha-srivastava: I will remove the test from here, you let the test be there.

s3tests_boto3/functional/__init__.py
TRYTOBE8TME: Okay

PR #401, opened 2021-07-14
Add test to check retain date is in iso 8601 format

dabukalam:
Signed-off-by: Danny Abukalam <danny@softiron.com>

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -12432,6 +12432,28 @@ def test_object_lock_get_obj_retention():
 
 @attr(resource='bucket')
 @attr(method='get')
+@attr(operation='Test object retention date formatting')
+@attr(assertion='success')
+@attr('object-lock')
+def test_object_lock_get_obj_retention_iso8601():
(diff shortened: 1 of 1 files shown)


cbodley:
@dabukalam can you please retarget this PR against the `master` branch instead of `ceph-master`? we can merge to master asap, then cherry-pick to ceph-master once the original fix merges

cbodley:
without the fix in https://github.com/ceph/ceph/pull/42273, the test fails with:
> isodate.isoerror.ISO8601Error: Unrecognised ISO 8601 date format: 'Fri, 15 Jul 2022 12:48:17 GM'

and passes with the fix 👍

alimaredia:
@dabukalam just so I understand the test correctly, if the test fails it would fail on the line `isodate.parse_datetime(retain_date)` correct?

dabukalam:
@alimaredia that's correct Ali, yes.

`parse_datetime()` will error out if it's not given a valid iso8601 date.

alimaredia (approved):

cbodley:
cherry-picked to ceph-master

PR #403, opened 2021-07-22
rgw sts abac tests

pritha-srivastava:
New STS tests for ABAC related changes.

### DIFF
--- a/s3tests.conf.SAMPLE
+++ b/s3tests.conf.SAMPLE
@@ -102,6 +102,8 @@ sub=<obtained after introspecting token>
 
 azp=<obtained after introspecting token>
 
+user_token=<access token for a user, with attribute Department=[Engineering, Marketing>]
+
 thumbprint=<obtained from x509 certificate>
 
 KC_REALM=<name of the realm>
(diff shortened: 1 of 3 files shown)


s3tests.conf.SAMPLE
```diff
@@ -92,6 +92,10 @@ token=<access_token>
 
 aud=<obtained after introspecting token>
 
+sub=<obtained after introspecting token>
+
+azp=<obtained after introspecting token>
+
```
TRYTOBE8TME, line 98: We should add `user_token` here

pritha-srivastava: yes i will add that later

pritha-srivastava:
Successful run of teuthology tests on branch https://github.com/ceph/ceph/pull/39849 is here: http://qa-proxy.ceph.com/teuthology/prsrivas-2021-07-30_09:49:12-rgw:sts-wip-rgw-sts-principal-tags-distro-basic-smithi/6303243/teuthology.log

alimaredia (approved):

PR #408, opened 2021-08-04
object-lock: test changes between retention modes

cbodley:
test cases for https://github.com/ceph/ceph/pull/42505

also reworks `nuke_prefixed_buckets()` to be aware of s3 object lock. some locked objects can't be deleted until their retention period expires, so each object lock test case should use a short retention period (~10 seconds), and `nuke_prefixed_buckets()` will wait up to 60 seconds to clean them up


cbodley:
thanks @alimaredia

PR #410, opened 2021-09-06
docs: fix wrong example in README.rst

Rjerk:
Run: S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests.functional.test_s3:test_bucket_list_empty

But get an error: "ERROR: Failure: ValueError (No such test test_bucket_list_empty)".

Because test_bucket_list_empty is a test case in s3tests_boto3 directory.

Signed-off-by: Liu Lan <liulan_yewu@cmss.chinamobile.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,15 +26,15 @@ Once you have that file copied and edited, you can run the tests with::
 
 You can specify which directory of tests to run::
 
-	S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests.functional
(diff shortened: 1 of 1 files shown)


alimaredia (approved):
This change is fine. 
It should be noted that not all of the tests in this suite are in the s3tests_boto3 directory . There are a few s3-tests in the `s3tests` dir that have not been ported over to boto3 yet for a handful of reasons.

PR #415, opened 2021-10-29
check Owner/Initiator fields of ListMultipartUploads response

cbodley:
new test case test_list_multipart_upload_owner() uses two different users to initiate multipart uploads, then tests that list_multipart_uploads() shows the correct user ids and display names for each upload


cbodley:
ping @alimaredia, the fix for this merged so it's safe to take this on master/ceph-master

cbodley:
ping @alimaredia

alimaredia (approved):

PR #380, opened 2021-01-11
Add testcases for rgw cloudtransition/cloudtiering feature

soumyakoduri:
Added few test cases for rgw cloud-transition/tiering feature. The sleep/waiting time in the tests may vary based on the I/O latency to the cloud endpoint.

Feature doc: https://docs.google.com/document/d/1IoeITPCF64A5W-UA-9Y3Vp2oSfz3xVQHu31GTu3u3Ug/edit#heading=h.4o1xyas8s23x

PR: ceph/ceph#35100

Also added existing boto2 lifecycle transition testcases to boto3

Signed-off-by: Soumya Koduri <skoduri@redhat.com>

PR #421, opened 2021-12-17
Enable Lifecycle tests

soumyakoduri:
This is to enable and execute lifecycle test cases in the teuthology run by adding an option to configure lc debug interval and adjust lifecycle tests sleep interval as per the value set.


soumyakoduri:
Note: These changes are on top of https://github.com/ceph/s3-tests/pull/380 .

soumyakoduri:
http://pulpito.front.sepia.ceph.com/soumyakoduri-2021-12-17_12:36:27-rgw:lifecycle-wip-skoduri-dbstore-mp-distro-basic-smithi

s3tests_boto3/functional/__init__.py
```diff
@@ -244,6 +247,12 @@ def setup():
         config.storage_classes = ""
         pass
 
+    try:
+        config.lc_debug_interval = int(cfg.get('s3 main',"lc_debug_interval"))
+    except (configparser.NoSectionError, configparser.NoOptionError):
+        config.lc_debug_interval = 10
+        pass
```
mattbenjamin, line 254: pass not needed if there is any statement on the branch?

s3tests_boto3/functional/test_s3.py
```diff
@@ -9240,7 +9248,9 @@ def test_lifecycle_expiration_tags1():
         Bucket=bucket_name, LifecycleConfiguration=lifecycle_config)
     eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
 
-    time.sleep(28)
+    lc_interval = get_lc_debug_interval()
+
+    time.sleep(3*lc_interval)
```
mattbenjamin, line 9253: yes, ok;

s3tests_boto3/functional/test_s3.py
```diff
@@ -9410,14 +9424,16 @@ def test_lifecycle_expiration_noncur_tags1():
...
     num_objs = verify_lifecycle_expiration_noncur_tags(
-        client, bucket_name, 20)
+        client, bucket_name, 2*lc_interval)
 
     # at T+20, 10 objects should exist
     eq(num_objs, 10)
 
     num_objs = verify_lifecycle_expiration_noncur_tags(
-        client, bucket_name, 40)
+        client, bucket_name, 5*lc_interval)
```
mattbenjamin, line 9436: My intuitions from prior experimentation was that assuming LC can schedule and expire in less than 1 debug interval, with interval 10s, is not always safe.

s3tests_boto3/functional/__init__.py
soumyakoduri: done

s3tests_boto3/functional/test_s3.py
soumyakoduri: I have adjusted the sleep interval to >=T+2 (T-debug interval) to account for any latency and also  increased the days count between successive lifecycle rules to accommodate that extra time. So far the tests have passed. 

I will trigger the test run few more times to confirm the same.

cbodley:
this contains everything from https://github.com/ceph/s3-tests/pull/380 right? so we can just merge this pr?

cbodley (approved):

soumyakoduri:
> this contains everything from #380 right? so we can just merge this pr?

yes. I will close the other PR then. Thanks!

cbodley:
ok, merged and cherry-picked to ceph-master (cc @alimaredia)

PR #422, opened 2021-12-22
Use a source-built nose instead of one in PyPI

zaitcev:
Fix crash on Fedora 35 with "AttributeError: module 'collections'
has no attribute 'Callable'". This happens because Callable is
in collections.abc in Python 3.10.

The culprit is PyPI, whence pip pulls the nose package. The nose
in Github works, the nose in Fedora works, but the wheel at PyPI
is broken. Someone hacked up the wheel that pip installs and
made a mistake. But with PyPI there's no recourse.

The maintainers disclaim any connection to the project anymore.

Our soultion is to switch to nose from the origin in github.
This may not be the best idea, because it downloads from
github on every test run.

Related-Tracker: https://tracker.ceph.com/issues/53478

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -74,3 +74,7 @@ ${virtualenv} --python=$(which python3.6) virtualenv
 # requirements.txt does not match setup.py requirements -- sucky but
 # good enough for now
 ./virtualenv/bin/python3 setup.py develop
+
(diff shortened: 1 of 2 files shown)


gaul:
Any resolution to this?  S3Proxy ran into this symptom:

https://github.com/gaul/s3proxy/pull/464#issuecomment-1363642389

PR #428, opened 2022-01-17
s3tests_boto3/test_sts: Changing code for proper cleanup

TRYTOBE8TME:
Signed-off-by: Kalpesh Pandya <kapandya@redhat.com>

### DIFF
--- a/s3tests_boto3/functional/test_sts.py
+++ b/s3tests_boto3/functional/test_sts.py
@@ -85,7 +85,7 @@ def put_user_policy(iam_client,username,policyname,policy_document):
         role_response = iam_client.put_user_policy(UserName=username,PolicyName=policyname,PolicyDocument=policy_document)
     except ClientError as e:
         role_err = e.response['Code']
-    return (role_err,role_response)
+    return (role_err,role_response,policyname)
 
 def get_s3_client_using_iam_creds():
     iam_access_key = get_iam_access_key()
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/test_sts.py
```diff
@@ -249,6 +251,7 @@ def test_assume_role_allow():
     eq(s3bucket['ResponseMetadata']['HTTPStatusCode'],200)
     bkt = s3_client.delete_bucket(Bucket=bucket_name)
     eq(bkt['ResponseMetadata']['HTTPStatusCode'],204)
+    iam_client.delete_role_policy(RoleName=general_role_name,PolicyName=policy_name)
```
pritha-srivastava, line 254: removing user policy for get session token makes sense. But I am not sure why deleting the role policy is needed for assume role tests, since the code clean up makes use of permanent credentials and role policy only affects temporary credentials (unless ofcourse there is a hidden bug). This needs to be debugged a bit more locally.

TRYTOBE8TME: Yes @pritha-srivastava you're correct. I guess we were seeing the error due to get_session_token tests (as per what I've checked).

s3tests_boto3/functional/test_sts.py
```diff
@@ -178,6 +178,7 @@ def test_get_session_token():
     s3bucket = s3_client.create_bucket(Bucket=bucket_name)
     eq(s3bucket['ResponseMetadata']['HTTPStatusCode'],200)
     finish=s3_client.delete_bucket(Bucket=bucket_name)
+    iam_client.delete_user_policy(UserName=sts_user_id,PolicyName=policy_name)
```
cbodley, line 181: can you please use try/finally for this cleanup?
```python
    try:
        s3bucket = s3_client.create_bucket(Bucket=bucket_name)
        eq(s3bucket['ResponseMetadata']['HTTPStatusCode'],200)
        finish=s3_client.delete_bucket(Bucket=bucket_name)
    finally: # clean up user policy even if create_bucket/delete_bucket fails
        iam_client.delete_user_policy(UserName=sts_user_id,PolicyName=policy_name)
```

cbodley (commented):
this looks like the right fix, based on my comment in https://tracker.ceph.com/issues/53090#note-5 👍

TRYTOBE8TME:
Passing teuthology: https://pulpito.ceph.com/kapandya-2022-01-18_06:29:54-rgw:sts-wip-rgw-sts-error-distro-basic-smithi/

pritha-srivastava (approved):

PR #437, opened 2022-04-05
Fix wrong assertion of the test: `test_buckets_list_ctime`

sumedhak27:
**TestName:**
s3tests_boto3.functional.test_s3:test_buckets_list_ctime

**Problem:**
The test creates 5 buckets for a user but in an assertion check, 
it asserts false if any bucket of the user has CreationTime less than a day prior to the current time.
Due to this reason, the test fails if the user has pre-existing buckets older than a day.

**Solution:**
Assert only on the CreationTime of buckets that were created with test execution.

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -6159,13 +6159,15 @@ def test_buckets_list_ctime():
     before = datetime.datetime.now(datetime.timezone.utc) - datetime.timedelta(days=1)
 
     client = get_client()
+    buckets = []
     for i in range(5):
-        client.create_bucket(Bucket=get_new_bucket_name())
+        buckets.append(client.create_bucket(Bucket=get_new_bucket_name()))
 
     response = client.list_buckets()
     for bucket in response['Buckets']:
-        ctime = bucket['CreationDate']
-        assert before <= ctime, '%r > %r' % (before, ctime)
+        if bucket['Name'] in buckets:
+            ctime = bucket['CreationDate']
+            assert before <= ctime, '%r > %r' % (before, ctime)
 
 @attr(resource='bucket')
 @attr(method='get')


LadMartin (approved):

cbodley (approved):

jheunis-bloomberg:
Is there something specific that is blocking this PR being merged? Or did it just slip through the cracks?
We've been running into this issue ourselves, would be handy if we could use this fix.

cbodley:
> Is there something specific that is blocking this PR being merged? Or did it just slip through the cracks? We've been running into this issue ourselves, would be handy if we could use this fix.

thanks for the reminder, @jheunis-bloomberg. @alimaredia would you be willing to test this?

bobham-bloomberg:
FYI, I cherry-picked this commit and tested it on our setup where it's ordinarily failing and it fixes the test:

```
2024-03-01 12:29:22,899 - DEBUG - 146 - Test OK after 0.73s: test_buckets_list_ctime
```

BBoozmen (approved):

BBoozmen:
> FYI, I cherry-picked this commit and tested it on our setup where the test is ordinarily failing and it fixes the test:
> 
> ```
> 2024-03-01 12:29:22,899 - DEBUG - 146 - Test OK after 0.73s: test_buckets_list_ctime
> ```

@cbodley - Bob has verified the fix. Can we merge this one in?

s3tests_boto3/functional/test_s3.py
```diff
@@ -6159,13 +6159,15 @@ def test_buckets_list_ctime():
     before = datetime.datetime.now(datetime.timezone.utc) - datetime.timedelta(days=1)
 
     client = get_client()
+    buckets = []
     for i in range(5):
-        client.create_bucket(Bucket=get_new_bucket_name())
+        buckets.append(client.create_bucket(Bucket=get_new_bucket_name()))
```
cbodley, line 6164: this creates an array of create_bucket() response objects, not of bucket names. so the loop below filters out all bucket names and doesn't check any ctimes

cbodley (changes requested):

s3tests_boto3/functional/test_s3.py
bobham-bloomberg: I've fixed this and created a new pull request, #565, as it looks like I don't have rights to update this pull request.

cbodley:
merged in https://github.com/ceph/s3-tests/pull/565, thanks @sumedhak27 @bobham-bloomberg

PR #446, opened 2022-04-27
Fix 400 or 403 which is always resolved to 400

LadMartin:


LadMartin:
Needs to be noted that all the test were previously failing on AWS because it returns 403.
We might need to force it to check for 403 only.

Issue #448, opened 2022-05-05
Planning to contribute to this project 


seagate-sarang-sawant:
Suites in this repository has compatibility and basic functional s3 protocol tests. My org is planning to contribute to this project in terms of functional and system tests using s3 protocol supported by Ceph RGW.  The aim to contribute only the functional tests with are agnostic of which backend is used while still following the Ceph's S3 protocol support. 

I am not sure if adding more functional tests(backend agnostic) is aligned with your purpose. 

ps: We would fork this repository and forked one will become our primary.

PR #453, opened 2022-05-19
Contributing User Policy tests

ketanarlulkar:
Inline with the issue created at #448 , adding few put/get/list/delete User Policy Tests.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -71,3 +71,31 @@ You can filter tests based on the attributes. There is a attribute named ``test_
 For running ``webidentity_test`` you'll need have Keycloak running.
 
 In order to run any STS test you'll need to add "iam" section to the config file. For further reference on how your config file should look check ``s3tests.conf.SAMPLE``.
+
+========================
+ IAM policy tests
+========================
+
+This is a set of IAM policy tests.
+This section covers tests for user policies such as Put, Get, List, Delete, user policies with s3 actions, conflicting user policies etc
+These tests uses Boto3 libraries. Tests are written in the ``s3test_boto3`` directory.
+
+These iam policy tests uses two users with profile name "iam" and "s3 alt" as mentioned in s3tests.conf.SAMPLE.
+If Ceph cluster is started with vstart, then above two users will get created as part of vstart with same access key, secrete key etc as mentioned in s3tests.conf.SAMPLE.
+Out of those two users, "iam" user is with capabilities --caps=user-policy=* and "s3 alt" user is without capabilities.
+Adding above capabilities to "iam" user is also taken care by vstart (If Ceph cluster is started with vstart).
+
+To run these tests, create configuration file with section "iam" and "s3 alt" refer s3tests.conf.SAMPLE.
+Once you have that configuration file copied and edited, you can run all the tests with::
+
+	S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests_boto3.functional.test_iam
+
+You can also specify specific test to run::
+
+	S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests_boto3.functional.test_iam:test_put_user_policy
+
+Some tests have attributes set such as "fails_on_rgw".
+You can filter tests based on their attributes::
+
+	S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests_boto3.functional.test_iam -a '!fails_on_rgw'
+
--- a/s3tests_boto3/functional/__init__.py
+++ b/s3tests_boto3/functional/__init__.py
@@ -435,6 +435,18 @@ def get_iam_client(client_config=None):
                         config=client_config)
     return client
 
+def get_iam_s3client(client_config=None):
+    if client_config == None:
+        client_config = Config(signature_version='s3v4')
+    client = boto3.client(service_name='s3',
+                          aws_access_key_id=get_iam_access_key(),
+                          aws_secret_access_key=get_iam_secret_key(),
+                          endpoint_url=config.default_endpoint,
+                          use_ssl=config.default_is_secure,
+                          verify=config.default_ssl_verify,
+                          config=client_config)
+    return client
+
 def get_alt_client(client_config=None):
     if client_config == None:
         client_config = Config(signature_version='s3v4')
@@ -479,6 +491,17 @@ def get_tenant_iam_client():
                           region_name='us-east-1',
                           aws_access_key_id=config.tenant_access_key,
                           aws_secret_access_key=config.tenant_secret_key,
+                          endpoint_url=config.default_endpoint,
+                          verify=config.default_ssl_verify,
+                          use_ssl=config.default_is_secure)
+    return client
+
+def get_alt_iam_client():
+
+    client = boto3.client(service_name='iam',
+                          region_name='',
+                          aws_access_key_id=config.alt_access_key,
+                          aws_secret_access_key=config.alt_secret_key,
                           endpoint_url=config.default_endpoint,
                           verify=config.default_ssl_verify,
                           use_ssl=config.default_is_secure)
--- a/s3tests_boto3/functional/test_iam.py
+++ b/s3tests_boto3/functional/test_iam.py
@@ -1,0 +1,963 @@
+import json
+
+from botocore.exceptions import ClientError
+from nose.plugins.attrib import attr
+from nose.tools import eq_ as eq
+
+from s3tests_boto3.functional.utils import assert_raises
+from s3tests_boto3.functional.test_s3 import _multipart_upload
+from . import (
+    get_alt_client,
+    get_iam_client,
+    get_new_bucket,
+    get_iam_s3client,
+    get_alt_iam_client,
+    get_alt_user_id,
+)
+from .utils import _get_status, _get_status_and_error_code
+
+
+@attr(resource='user-policy')
+@attr(method='put')
+@attr(operation='Verify Put User Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_put_user_policy():
+    client = get_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+    response = client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
+                                      UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    response = client.delete_user_policy(PolicyName='AllAccessPolicy',
+                                         UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+
+
+@attr(resource='user-policy')
+@attr(method='put')
+@attr(operation='Verify Put User Policy with invalid user')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_put_user_policy_invalid_user():
+    client = get_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+    e = assert_raises(ClientError, client.put_user_policy, PolicyDocument=policy_document,
+                      PolicyName='AllAccessPolicy', UserName="some-non-existing-user-id")
+    status = _get_status(e.response)
+    eq(status, 404)
+
+
+@attr(resource='user-policy')
+@attr(method='put')
+@attr(operation='Verify Put User Policy using parameter value outside limit')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_put_user_policy_parameter_limit():
+    client = get_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": [{
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}] * 1000
+         }
+    )
+    e = assert_raises(ClientError, client.put_user_policy, PolicyDocument=policy_document,
+                      PolicyName='AllAccessPolicy' * 10, UserName=get_alt_user_id())
+    status = _get_status(e.response)
+    eq(status, 400)
+
+
+@attr(resource='user-policy')
+@attr(method='put')
+@attr(operation='Verify Put User Policy using invalid policy document elements')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+@attr('fails_on_rgw')
+def test_put_user_policy_invalid_element():
+    client = get_iam_client()
+
+    # With Version other than 2012-10-17
+    policy_document = json.dumps(
+        {"Version": "2010-10-17",
+         "Statement": [{
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}]
+         }
+    )
+    e = assert_raises(ClientError, client.put_user_policy, PolicyDocument=policy_document,
+                      PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+    status = _get_status(e.response)
+    eq(status, 400)
+
+    # With no Statement
+    policy_document = json.dumps(
+        {
+            "Version": "2012-10-17",
+        }
+    )
+    e = assert_raises(ClientError, client.put_user_policy, PolicyDocument=policy_document,
+                      PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+    status = _get_status(e.response)
+    eq(status, 400)
+
+    # with same Sid for 2 statements
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": [
+             {"Sid": "98AB54CF",
+              "Effect": "Allow",
+              "Action": "*",
+              "Resource": "*"},
+             {"Sid": "98AB54CF",
+              "Effect": "Allow",
+              "Action": "*",
+              "Resource": "*"}]
+         }
+    )
+    e = assert_raises(ClientError, client.put_user_policy, PolicyDocument=policy_document,
+                      PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+    status = _get_status(e.response)
+    eq(status, 400)
+
+    # with Principal
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": [{
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*",
+             "Principal": "arn:aws:iam:::username"}]
+         }
+    )
+    e = assert_raises(ClientError, client.put_user_policy, PolicyDocument=policy_document,
+                      PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+    status = _get_status(e.response)
+    eq(status, 400)
+
+
+@attr(resource='user-policy')
+@attr(method='put')
+@attr(operation='Verify Put a policy that already exists')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_put_existing_user_policy():
+    client = get_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}
+         }
+    )
+    response = client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
+                                      UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
+                           UserName=get_alt_user_id())
+    client.delete_user_policy(PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+
+
+@attr(resource='user-policy')
+@attr(method='put')
+@attr(operation='Verify List User policies')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_list_user_policy():
+    client = get_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}
+         }
+    )
+    response = client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
+                                      UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    response = client.list_user_policies(UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    client.delete_user_policy(PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+
+
+@attr(resource='user-policy')
+@attr(method='put')
+@attr(operation='Verify List User policies with invalid user')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_list_user_policy_invalid_user():
+    client = get_iam_client()
+    e = assert_raises(ClientError, client.list_user_policies, UserName="some-non-existing-user-id")
+    status = _get_status(e.response)
+    eq(status, 404)
+
+
+@attr(resource='user-policy')
+@attr(method='get')
+@attr(operation='Verify Get User policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_get_user_policy():
+    client = get_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+    response = client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
+                                      UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    response = client.get_user_policy(PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+
+    response = client.delete_user_policy(PolicyName='AllAccessPolicy',
+                                         UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+
+
+@attr(resource='user-policy')
+@attr(method='get')
+@attr(operation='Verify Get User Policy with invalid user')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_get_user_policy_invalid_user():
+    client = get_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+    response = client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
+                                      UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    e = assert_raises(ClientError, client.get_user_policy, PolicyName='AllAccessPolicy',
+                      UserName="some-non-existing-user-id")
+    status = _get_status(e.response)
+    eq(status, 404)
+    client.delete_user_policy(PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+
+
+@attr(resource='user-policy')
+@attr(method='get')
+@attr(operation='Verify Get User Policy with invalid policy name')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+@attr('fails_on_rgw')
+def test_get_user_policy_invalid_policy_name():
+    client = get_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+    client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
+                           UserName=get_alt_user_id())
+    e = assert_raises(ClientError, client.get_user_policy, PolicyName='non-existing-policy-name',
+                      UserName=get_alt_user_id())
+    status = _get_status(e.response)
+    eq(status, 404)
+    client.delete_user_policy(PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+
+
+@attr(resource='user-policy')
+@attr(method='get')
+@attr(operation='Verify Get Deleted User Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+@attr('fails_on_rgw')
+def test_get_deleted_user_policy():
+    client = get_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+    client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
+                           UserName=get_alt_user_id())
+    client.delete_user_policy(PolicyName='AllAccessPolicy', UserName=get_alt_user_id())
+    e = assert_raises(ClientError, client.get_user_policy, PolicyName='AllAccessPolicy',
+                      UserName=get_alt_user_id())
+    status = _get_status(e.response)
+    eq(status, 404)
+
+
+@attr(resource='user-policy')
+@attr(method='get')
+@attr(operation='Verify Get a policy from multiple policies for a user')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_get_user_policy_from_multiple_policies():
+    client = get_iam_client()
+
+    policy_document_allow = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+
+    response = client.put_user_policy(PolicyDocument=policy_document_allow,
+                                      PolicyName='AllowAccessPolicy1',
+                                      UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    response = client.put_user_policy(PolicyDocument=policy_document_allow,
+                                      PolicyName='AllowAccessPolicy2',
+                                      UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    response = client.get_user_policy(PolicyName='AllowAccessPolicy2',
+                                      UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+
+    response = client.delete_user_policy(PolicyName='AllowAccessPolicy1',
+                                         UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    response = client.delete_user_policy(PolicyName='AllowAccessPolicy2',
+                                         UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+
+
+@attr(resource='user-policy')
+@attr(method='delete')
+@attr(operation='Verify Delete User Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_delete_user_policy():
+    client = get_iam_client()
+
+    policy_document_allow = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+
+    response = client.put_user_policy(PolicyDocument=policy_document_allow,
+                                      PolicyName='AllowAccessPolicy',
+                                      UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    response = client.delete_user_policy(PolicyName='AllowAccessPolicy',
+                                         UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+
+
+@attr(resource='user-policy')
+@attr(method='delete')
+@attr(operation='Verify Delete User Policy with invalid user')
+@attr(assertion='succeeds')
+@attr('user-policy')
+@attr('test_of_iam')
+def test_delete_user_policy_invalid_user():
+    client = get_iam_client()
+
+    policy_document_allow = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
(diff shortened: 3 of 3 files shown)


alimaredia:
@ketanarlulkar @ravindrachoudhari @seagate-sarang-sawant could you squash all the commits down to one commit (or a lower number) so that this code is easier to cherry-pick once you're all set. Also if I were to test this on a Ceph cluster are there any Ceph config variables I would need to configure?

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,218 @@
...
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+    client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
+                           UserName=get_tenant_user_id())
```
pritha-srivastava, line 34: PutUserPolicy is used to set a user policy by a user with user-policy caps on a different user. I believe the access key, secret access key that is being used to create iam_client is the same for the user returned by get_tenant_user_id()

pritha-srivastava: Also, what does this test check for? Success?. There is a utility method for put_user_policy in https://github.com/ceph/s3-tests/blob/master/s3tests_boto3/functional/test_sts.py#L80, that could be put in a common file and re-used for all put user policy tests.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,218 @@
...
+    client = get_tenant_iam_client()
+
+    policy_document = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": "*",
+             "Resource": "*"}}
+    )
+    client.put_user_policy(PolicyDocument=policy_document, PolicyName='AllAccessPolicy',
```
pritha-srivastava, line 165: How is it being verified that the call succeeds?

seagate-sarang-sawant:
> @ketanarlulkar @ravindrachoudhari @seagate-sarang-sawant could you squash all the commits down to one commit (or a lower number) so that this code is easier to cherry-pick once you're all set. Also if I were to test this on a Ceph cluster are there any Ceph config variables I would need to configure?

Yes it was a mistake to not keep `squash and merge` as default in our fork. We have made 'squash and merge' default recently. Let us fix the PR.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,218 @@
...
```
pritha-srivastava, line 188: Generic comment - please check for success wherever applicable.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,331 @@
...
+                           UserName=get_tenant_user_id())
+
+    client.get_user_policy(PolicyName='AllAccessPolicy', UserName=get_tenant_user_id())
```
pritha-srivastava, line 224: Please check if the correct policy is returned.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,331 @@
...
+             "Resource": "*"}}
+    )
+
+    client.put_user_policy(PolicyDocument=policy_document_allow, PolicyName='AllowAccessPolicy1',
+                           UserName=get_tenant_user_id())
+
+    client.put_user_policy(PolicyDocument=policy_document_allow, PolicyName='AllowAccessPolicy2',
+                           UserName=get_tenant_user_id())
+
+    client.get_user_policy(PolicyName='AllowAccessPolicy2', UserName=get_tenant_user_id())
```
pritha-srivastava: same comment - check if the policy returned is the one that is being expected.

s3tests_boto3/functional/test_iam.py
```diff
@@ -329,3 +329,112 @@ def test_get_deleted_user_policy():
...
+
+    client.put_user_policy(PolicyDocument=policy_document_allow, PolicyName='AllowAccessPolicy',
+                           UserName=get_tenant_user_id())
+
+    client.delete_user_policy(PolicyName='AllowAccessPolicy', UserName=get_tenant_user_id())
+
+    e = assert_raises(ClientError, client.get_user_policy, PolicyName='AllAccessPolicy',
+                      UserName=get_tenant_user_id())
+    status = _get_status(e.response)
+    eq(status, 404)
```
pritha-srivastava: is this a success case? in that case will the return code be 404?

s3tests_boto3/functional/test_iam.py
```diff
@@ -329,3 +329,112 @@ def test_get_deleted_user_policy():
...
+                           UserName=get_tenant_user_id())
+
+    client.put_user_policy(PolicyDocument=policy_document_allow, PolicyName='AllowAccessPolicy3',
+                           UserName=get_tenant_user_id())
+
+    client.delete_user_policy(PolicyName='AllowAccessPolicy1', UserName=get_tenant_user_id())
+
+    client.delete_user_policy(PolicyName='AllowAccessPolicy2', UserName=get_tenant_user_id())
+
+    client.get_user_policy(PolicyName='AllowAccessPolicy3', UserName=get_tenant_user_id())
```
pritha-srivastava, line 420: Also please check if 'AllowAccessPolicy1' and 'AllowAccessPolicy2' have been deleted by either using ListUserPolicies or GetUserPolicy.

s3tests_boto3/functional/test_iam.py
```diff
@@ -413,3 +419,300 @@ def test_delete_user_policy_from_multiple_policies():
...
+    policy_document_allow = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Allow",
+             "Action": ["s3:ListAllMyBuckets", "s3:DeleteBucket"],
+             "Resource": "arn:aws:s3:::*"}}
+    )
+    client.put_user_policy(PolicyDocument=policy_document_allow, PolicyName='AllowAccessPolicy',
+                           UserName=get_tenant_user_id())
+    s3_client = get_client()
```
pritha-srivastava, line 443: 1. Please attach user policy to a different user using the iam_client.
2. s3_client is not created using the same user_id to which the user policy is attached (correct me if I am wrong), so is it being really tested if the user policy is exercised?
3. In order to create s3_client using creds of the user to which the user policy is attached, please see: https://github.com/ceph/s3-tests/blob/master/s3tests_boto3/functional/test_sts.py#L90

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: This is not a success case. we are deleting the user policy and then trying to get the user policy.
In that case it will return 404.

pritha-srivastava:
> @ketanarlulkar @ravindrachoudhari @seagate-sarang-sawant could you squash all the commits down to one commit (or a lower number) so that this code is easier to cherry-pick once you're all set. Also if I were to test this on a Ceph cluster are there any Ceph config variables I would need to configure?

I agree, squashing comments will make it easier to review.

s3tests_boto3/functional/test_iam.py
pritha-srivastava: ok.

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: If it fails then boto API will throw the exception and test will fail there itself.

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: Yes, we will add the check.

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: ok, we will add the check.

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: ok, we will add the check.

s3tests_boto3/functional/test_iam.py
pritha-srivastava: Yes, but its easier for someone reading the tests also to understand as to what is the expected outcome.

ravindrachoudhari: Ok, we will add the check

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: Ok, we will add the check

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: yes, we will use the same user(on which user policy is applied) to create s3_client.

s3tests_boto3/functional/test_iam.py
```diff
@@ -413,3 +419,300 @@ def test_delete_user_policy_from_multiple_policies():
...
+        {"Version": "2012-10-17",
+         "Statement": {
+             "Effect": "Deny",
+             "Action": ["s3:ListAllMyBuckets", "s3:DeleteBucket"],
+             "Resource": "arn:aws:s3:::*"}}
+    )
+
+    client.put_user_policy(PolicyDocument=policy_document_deny, PolicyName='DenyAccessPolicy',
+                           UserName=get_tenant_user_id())
+    s3_client = get_client()
```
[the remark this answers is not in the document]
ravindrachoudhari: yes, we will use the same user(on which user policy is applied) to create s3_client.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,934 @@
...
+    )
+    e = assert_raises(ClientError, client.put_user_policy, PolicyDocument=policy_document,
+                      PolicyName='AllAccessPolicy' * 10, UserName=get_tenant_user_id())
+    status = _get_status(e.response)
+    eq(status, 400)
+
+
+@attr(resource='user-policy')
+@attr(method='put')
+@attr(operation='Verify Put User Policy using invalid policy document elements')
```
pritha-srivastava, line 86: It can be mentioned here that version is invalid.

pritha-srivastava: nevermind, its there in the comment.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,934 @@
...
+
+@attr(resource='user-policy')
+@attr(method='s3 Actions')
+@attr(operation='Verify Allow Bucket Actions in user Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+def test_allow_bucket_actions_in_user_policy():
+    client = get_iam_client()
+    s3_client = get_tenant_client()
+    bucket = get_new_bucket(client=s3_client)
```
pritha-srivastava, line 491: here get_new_bucket() gets bucket for the tenant client and then ListBuckets and DeleteBucket is being tested. These actions will be allowed even without the user policy since tenant client is the owner of the bucket.
The bucket must be created for another user in the same tenant and then the policy must be attached to the tenant user and then the policy must be exercised to see whether the tenant user has ListAllMyBuckets and DeleteBucket permission over buckets belonging to the other user.

pritha-srivastava: Or if you want to work with the bucket owned by the tenant user, you can modify the policy to only allow Read Access to the bucket and the objects in it. And check if write to the bucket is being 'Denied' correctly.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,934 @@
...
+
+@attr(resource='user-policy')
+@attr(method='s3 Actions')
+@attr(operation='Verify Deny Bucket Actions in user Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+def test_deny_bucket_actions_in_user_policy():
+    client = get_iam_client()
+    s3_client = get_tenant_client()
+    bucket = get_new_bucket(client=s3_client)
```
pritha-srivastava, line 537: this test is ok, because the bucket being owned by tenant user, has all s3 actions allowed on it by default. The policy explicitly 'Denies' two actions and that is being tested here.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,934 @@
...
+@attr(method='s3 Actions')
+@attr(operation='Verify Allow Object Actions in user Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+def test_allow_object_actions_in_user_policy():
+    client = get_iam_client()
+    s3_client = get_tenant_client()
+    bucket = get_new_bucket(client=s3_client)
+
+    policy_document_allow = json.dumps(
```
pritha-srivastava, line 577: same comment here - the bucket belongs to tenant user, so all s3 actions will be allowed by default, even if you dont attach this policy.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,934 @@
...
+@attr(operation='Verify Deny Object Actions in user Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+def test_deny_object_actions_in_user_policy():
+    client = get_iam_client()
+    s3_client = get_tenant_client()
+    bucket = get_new_bucket(client=s3_client)
+
+    policy_document_deny = json.dumps(
+        {"Version": "2012-10-17",
```
pritha-srivastava, line 618: this is ok.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,934 @@
...
+@attr(assertion='succeeds')
+@attr('user-policy')
+def test_allow_multipart_actions_in_user_policy():
+    client = get_iam_client()
+    s3_client = get_tenant_client()
+    bucket = get_new_bucket(client=s3_client)
+
+    policy_document_allow = json.dumps(
+        {"Version": "2012-10-17",
+         "Statement": {
```
pritha-srivastava, line 661: Same comment for all 'Allow' tests.

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: Yes, you are right.
we will do the changes.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,939 @@
...
+@attr(resource='user-policy')
+@attr(method='s3 Actions')
+@attr(operation='Verify Allow Bucket Actions in user Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+def test_allow_bucket_actions_in_user_policy():
+    client = get_iam_client()
+    s3_client_tenant = get_tenant_client()
+
+    s3_client_iam = get_iam_s3client()
```
pritha-srivastava, line 488: Here again the bucket is created by the iam_user (via s3 api), and the policy is attached to the iam_user that is creating the bucket right? and then the bucket is being accessed by a user in a different tenant. Has this been tested against RGW?

pritha-srivastava: no, sorry the policy is attached to the tenant user. And now the tenant user tries to access the bucket which is in the tenant of the iam user and these two tenants are different right? Does this pass against RGW?

pritha-srivastava: > 1. policy

You may have to fix these discrepancies in RGW.

pritha-srivastava: > Yes, this passes against RGW.
> 
> But we are seeing some issues with RGW
> 
> 1. when we try to get non existing policy (or deleted policy) it return malformed response which boto3 is unable to decode where as AWS returns "NoSuchEntity"
> 2. Policy without statement gets accepted where as AWS returns error "MalformedPolicyDocument"
> 3. Policy with two same Statement IDs in a single policy gets accepted where as AWS returns " MalformedPolicyDocument"

For all three, you will have to fix the code.
> 
> Due to above three issues, three tests are failing on RGW
> 
> 1. test_put_user_policy_invalid_element
> 2. test_get_user_policy_invalid_policy_name
> 3. test_get_deleted_user_policy

pritha-srivastava: > Are you sure that this passes? because it doesnt for me on the latest master. I have the exact same test where I create an iam_client in 'empty' tenant and attach a policy to a user in a tenant 'testx'. Then I create a bucket and upload an object accessing the 's3' api using the iam_client. Then I create an s3 connection using the user in tenant 'testx', and try to list objects - ListBucket fails with error - Bucket not found and rightly so, since when the bucket you have created using iam_client is under 'empty' tenant, and when the tenant user tries to access it, the bucket is searched in 'testx' tenant and this fails. The following are the logs in rgw, for list_bucket: 2022-06-03T20:46:27.660+0530 7f3a61024700 10 req 3483647770686675559 0.001000012s s->object= s->bucket=testx/my-new-shiny-bucket 2022-06-03T20:46:27.660+0530 7f3a61024700 2 req 3483647770686675559 0.001000012s s3:list_bucket init permissions 2022-06-03T20:46:27.660+0530 7f3a61024700 20 req 3483647770686675559 0.001000012s **s3:list_bucket get_system_obj_state: rctx=0x7f39ec6372b0 obj=default.rgw.meta:root:testx/my-new-shiny-bucket state=0x55801ad1e0a0 s->prefetch_data=0 2022-06-03T20:46:27.660+0530 7f3a61024700 10 req 3483647770686675559 0.001000012s** s3:list_bucket cache get: name=default.rgw.meta+root+testx/my-new-shiny-bucket : miss ... **2022-06-03T20:46:27.661+0530 7f3a5d81d700 5 bucket my-new-shiny-bucket doesn't exist, not logging**
> 
> Can you verify that the tenants of the iam_client and the tenanted_user are the same or not, if they are then this is a valid test and it must pass, else it must not.

Just for your reference, I am attaching the rgw logs for create_bucket also here:
2022-06-03T20:42:26.829+0530 7f3aaf8c1700  2 req 10478071630861750662 0.003000033s s3:create_bucket normalizing buckets and tenants
**2022-06-03T20:42:26.829+0530 7f3aaf8c1700 10 req 10478071630861750662 0.003000033s s->object=<NULL> s->bucket=my-new-shiny-bucket**
2022-06-03T20:42:26.831+0530 7f3aae0be700 10 req 10478071630861750662 0.005000055s s3:create_bucket cache put: name=default.rgw.meta+root+my-new-shiny-bucket info.flags=0x0
**2022-06-03T20:42:26.831+0530 7f3aae0be700 10 req 10478071630861750662 0.005000055s s3:create_bucket adding default.rgw.meta+root+my-new-shiny-bucket to cache LRU end**
2022-06-03T20:42:26.831+0530 7f3aae0be700 10 req 10478071630861750662 0.005000055s s3:create_bucket user=TESTER bucket=:my-new-shiny-bucket[])

You can see that in create_bucket there is no tenant: default.rgw.meta+root+my-new-shiny-bucket
Whereas, in list_bucket it searches for the bucket under tenant 'testx': default.rgw.meta:root:testx/my-new-shiny-bucket
Hence the search fails. This just for you to verify what the test is actually doing.

ravindrachoudhari: Yes, both the users were in same tenant and now using s3 alt and iam user which are supposed to be under same tenant.
Also fixed the code for failing tests.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,857 @@
...
+
+@attr(resource='user-policy')
+@attr(method='s3 Actions')
+@attr(operation='Verify Deny Bucket Actions in user Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+def test_deny_bucket_actions_in_user_policy():
+    client = get_iam_client()
+    s3_client = get_alt_client()
+    bucket = get_new_bucket(client=s3_client)
```
pritha-srivastava, line 455: shouldnt the bucket be created by s3_client_iam like in the previous test?

pritha-srivastava: or are you trying to test Deny on the bucket owned by alt_client?

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,857 @@
...
+@attr(resource='user-policy')
+@attr(method='s3 Actions')
+@attr(operation='Verify Deny Object Actions in user Policy')
+@attr(assertion='succeeds')
+@attr('user-policy')
+def test_deny_object_actions_in_user_policy():
+    client = get_iam_client()
+    s3_client_alt = get_alt_client()
+    s3_client_iam = get_iam_s3client()
+    bucket = get_new_bucket(client=s3_client_iam)
```
pritha-srivastava, line 536: actually 'Deny' cases must be tested on the bucket owned by alt_client itself, since by default all s3 Actions will be 'Denied' on a bucket belonging to a different user. It should be similar to the previous 'Deny' case that has been written.

s3tests_boto3/functional/test_iam.py
```diff
@@ -0,0 +1,857 @@
...
+                                  "iam:ListUserPolicies", "iam:DeleteUserPolicy"],
+                       "Resource": f"arn:aws:iam:::user/{get_alt_user_id()}"}}
+    )
+    client1 = get_iam_client()
+    iam_client_tenant = get_tenant_iam_client()
+
+    response = client1.put_user_policy(PolicyDocument=policy1, PolicyName='AllowAccessPolicy',
+                                       UserName=get_alt_user_id())
+    eq(response['ResponseMetadata']['HTTPStatusCode'], 200)
+    response = iam_client_tenant.get_user_policy(PolicyName='AllowAccessPolicy',
```
pritha-srivastava, line 850: shouldn't the client be corresponding to the alt_user since the policy is being attached to alt_user?

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: As you mentioned in one of your comment 

`this test is ok, because the bucket being owned by tenant user, has all s3 actions allowed on it by default. The policy explicitly 'Denies' two actions and that is being tested here.`

So we have not made any changes in any deny tests

pritha-srivastava: Yes this makes sense.

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: yes, will fix this

s3tests_boto3/functional/test_iam.py
ravindrachoudhari: yes, done

pritha-srivastava:
The tests look fine now. I am not sure about commits like -Merge pull request from ..., @alimaredia : can you comment on this?

ketanarlulkar:
Squashed all commits into single commit.

ravindrachoudhari:
> @ketanarlulkar @ravindrachoudhari @seagate-sarang-sawant could you squash all the commits down to one commit (or a lower number) so that this code is easier to cherry-pick once you're all set. Also if I were to test this on a Ceph cluster are there any Ceph config variables I would need to configure?

@alimaredia squash is done. Please review.
No need to configure any Ceph variables.
We have deployed Ceph with default config.
For test execution, only s3test.conf file needs to be updated with user info(access key, secret key, user id etc)

ravindrachoudhari (approved):

s3tests_boto3/functional/__init__.py
```diff
@@ -484,6 +496,17 @@ def get_tenant_iam_client():
                           use_ssl=config.default_is_secure)
     return client
 
+def get_alt_iam_client():
+
+    client = boto3.client(service_name='iam',
+                          region_name='us-east-1',
```
seagate-sarang-sawant, line 502: I see region as empty string in few functions.

seagate-sarang-sawant (approved):

pritha-srivastava:
I ran the tests locally, and all tests passed. One thing that I had to do was to add --caps=user-policy=* to the iam user in vstart.sh @alimaredia : should this be added to vstart.sh? or someone running these tests is expected to manually add this caps to the iam_user in their config file? Also I think this change will be required in teuthology (adding user-policy caps to iam user). 
@ketanarlulkar @ravindrachoudhari : You said some tests were failing because of mismatch in the expected error code by AWS and that returned by RGW and that you modified the code. But all tests passed for me? Can you explain how is this possible?
Readme needs to be updated for User Policy/IAM tests.

ravindrachoudhari:
> I ran the tests locally, and all tests passed. One thing that I had to do was to add --caps=user-policy=* to the iam user in vstart.sh @alimaredia : should this be added to vstart.sh? or someone running these tests is expected to manually add this caps to the iam_user in their config file? Also I think this change will be required in teuthology (adding user-policy caps to iam user). @ketanarlulkar @ravindrachoudhari : You said some tests were failing because of mismatch in the expected error code by AWS and that returned by RGW and that you modified the code. But all tests passed for me? Can you explain how is this possible? Readme needs to be updated for User Policy/IAM tests.

@pritha-srivastava Thanks for trying out the tests.
Yes, I forgot to mention about the caps.
About the failing tests, out of 3 failing tests one we fixed and 2 tests currently we removed as there is some issue with boto3 (unable to decode the response.)

ravindrachoudhari:
> 
when we try to get non existing policy (or deleted policy) it return malformed response which boto3 is unable to decode

pritha-srivastava:
> > I ran the tests locally, and all tests passed. One thing that I had to do was to add --caps=user-policy=* to the iam user in vstart.sh @alimaredia : should this be added to vstart.sh? or someone running these tests is expected to manually add this caps to the iam_user in their config file? Also I think this change will be required in teuthology (adding user-policy caps to iam user). @ketanarlulkar @ravindrachoudhari : You said some tests were failing because of mismatch in the expected error code by AWS and that returned by RGW and that you modified the code. But all tests passed for me? Can you explain how is this possible? Readme needs to be updated for User Policy/IAM tests.
> 
> @pritha-srivastava Thanks for trying out the tests. Yes, I forgot to mention about the caps. About the failing tests, out of 3 failing tests one we fixed and 2 tests currently we removed as there is some issue with boto3 (unable to decode the response.)

How did you fix the one that you fixed? Did you modify the test itself

ravindrachoudhari:
> > > I ran the tests locally, and all tests passed. One thing that I had to do was to add --caps=user-policy=* to the iam user in vstart.sh @alimaredia : should this be added to vstart.sh? or someone running these tests is expected to manually add this caps to the iam_user in their config file? Also I think this change will be required in teuthology (adding user-policy caps to iam user). @ketanarlulkar @ravindrachoudhari : You said some tests were failing because of mismatch in the expected error code by AWS and that returned by RGW and that you modified the code. But all tests passed for me? Can you explain how is this possible? Readme needs to be updated for User Policy/IAM tests.
> > 
> > 
> > @pritha-srivastava Thanks for trying out the tests. Yes, I forgot to mention about the caps. About the failing tests, out of 3 failing tests one we fixed and 2 tests currently we removed as there is some issue with boto3 (unable to decode the response.)
> 
> How did you fix the one that you fixed? Did you modify the test itself
Yes

alimaredia:
@pritha-srivastava a teuthology task related change should merge as the same time as this PR. I think that the `--caps=user-policy=*` should be added by the user manually, but I'd like to discuss adding that in the Ceph PR with the teuthology changes. 

I'd like most s3-tests to pass with a vstart cluster out of the box with the config file provided in the s3-tests repo but I don't think that's the case right now. Nevertheless, there's documentation that includes adding that flag when modifying or create an RGW user right?

ravindrachoudhari:
> @pritha-srivastava a teuthology task related change should merge as the same time as this PR. I think that the `--caps=user-policy=*` should be added by the user manually, but I'd like to discuss adding that in the Ceph PR with the teuthology changes.
> 
> I'd like most s3-tests to pass with a vstart cluster out of the box with the config file provided in the s3-tests repo but I don't think that's the case right now. Nevertheless, there's documentation that includes adding that flag when modifying or create an RGW user right?

@alimaredia @pritha-srivastava 
Sorry I am not able to make any AI out of it for us. Any changes are needed in this PR?

pritha-srivastava:
> @pritha-srivastava a teuthology task related change should merge as the same time as this PR. I think that the `--caps=user-policy=*` should be added by the user manually, but I'd like to discuss adding that in the Ceph PR with the teuthology changes.
> 
> I'd like most s3-tests to pass with a vstart cluster out of the box with the config file provided in the s3-tests repo but I don't think that's the case right now. Nevertheless, there's documentation that includes adding that flag when modifying or create an RGW user right?

@alimaredia : the documentation for adding caps for user-policy is here: https://docs.ceph.com/en/quincy/radosgw/STSLite/
Also, if we want to add these changes in the Ceph PR then I think we need to mention in the README that a certain user is created in the vstart cluster with user-policy caps and must be used for running tests locally.

pritha-srivastava:
> > > > I ran the tests locally, and all tests passed. One thing that I had to do was to add --caps=user-policy=* to the iam user in vstart.sh @alimaredia : should this be added to vstart.sh? or someone running these tests is expected to manually add this caps to the iam_user in their config file? Also I think this change will be required in teuthology (adding user-policy caps to iam user). @ketanarlulkar @ravindrachoudhari : You said some tests were failing because of mismatch in the expected error code by AWS and that returned by RGW and that you modified the code. But all tests passed for me? Can you explain how is this possible? Readme needs to be updated for User Policy/IAM tests.
> > > 
> > > 
> > > @pritha-srivastava Thanks for trying out the tests. Yes, I forgot to mention about the caps. About the failing tests, out of 3 failing tests one we fixed and 2 tests currently we removed as there is some issue with boto3 (unable to decode the response.)
> > 
> > 
> > How did you fix the one that you fixed? Did you modify the test itself
> > Yes

The correct way to do this would be open a RGW PR with the corrections needed, and till then add an attr @fails_on_rgw to the tests that fail currently, and when the RGW PR gets merged then the attr can be removed. @alimaredia : what do you say?

pritha-srivastava:
When I check the s3tests.py, I can see that user-policy caps have already been added to the 'iam' user (this was done as part of STS tests by Kalpesh) https://github.com/ceph/ceph/blob/main/qa/tasks/s3tests.py#L112. So a teuthology run can also be scheduled for this PR once all the changes are final.

ravindrachoudhari:
> > > > > I ran the tests locally, and all tests passed. One thing that I had to do was to add --caps=user-policy=* to the iam user in vstart.sh @alimaredia : should this be added to vstart.sh? or someone running these tests is expected to manually add this caps to the iam_user in their config file? Also I think this change will be required in teuthology (adding user-policy caps to iam user). @ketanarlulkar @ravindrachoudhari : You said some tests were failing because of mismatch in the expected error code by AWS and that returned by RGW and that you modified the code. But all tests passed for me? Can you explain how is this possible? Readme needs to be updated for User Policy/IAM tests.
> > > > 
> > > > 
> > > > @pritha-srivastava Thanks for trying out the tests. Yes, I forgot to mention about the caps. About the failing tests, out of 3 failing tests one we fixed and 2 tests currently we removed as there is some issue with boto3 (unable to decode the response.)
> > > 
> > > 
> > > How did you fix the one that you fixed? Did you modify the test itself
> > > Yes
> 
> The correct way to do this would be open a RGW PR with the corrections needed, and till then add an attr @fails_on_rgw to the tests that fail currently, and when the RGW PR gets merged then the attr can be removed. @alimaredia : what do you say?

@pritha-srivastava we have raised bug for these failures against our Seagate RGW dev team, right now I can not comment on the PR's ETA.
We are working on some more scenarios and tests for user policies, once this PR gets merged we can add those failed tests in the next PR with attr @fails_on_rgw if you are ok with that.

mattbenjamin:
@ravindrachoudhari I'm not sure why we wouldn't establish the right pattern (e.g., @fails_on_rgw) for the tests that we already have in the PR.  It seems preferable to waiting to "fix everything up" in a later one with no definite arrival date, as above...

ravindrachoudhari:
> @ravindrachoudhari I'm not sure why we wouldn't establish the right pattern (e.g., @fails_on_rgw) for the tests that we already have in the PR. It seems preferable to waiting to "fix everything up" in a later one with no definite arrival date, as above...

@mattbenjamin The ETA I am not sure about is from dev team, not for adding tests. we are planning to start new PR within a week for next set of test cases.
But as you rightly mentioned we should establish right pattern, we will add those failing test cases back in this PR only with attr @fails_on_rgw.

ravindrachoudhari:
@mattbenjamin @pritha-srivastava @alimaredia 
Added those failing three tests tests back with attr @fails_on_rgw.
Can you please review and merge(if no more changes are required)

pritha-srivastava:
> I ran the tests locally, and all tests passed. One thing that I had to do was to add --caps=user-policy=* to the iam user in vstart.sh @alimaredia : should this be added to vstart.sh? or someone running these tests is expected to manually add this caps to the iam_user in their config file? Also I think this change will be required in teuthology (adding user-policy caps to iam user). @ketanarlulkar @ravindrachoudhari : You said some tests were failing because of mismatch in the expected error code by AWS and that returned by RGW and that you modified the code. But all tests passed for me? Can you explain how is this possible? Readme needs to be updated for User Policy/IAM tests.

The following three are still unresolved:

1. The README hasn't been updated.
2. A clean teuthology run.
3. We still havent decided whether vstart.sh will have user-policy caps added for the iam user or is the user expected to do this manually. If the user is expected to do this manually, then that needs to go into the README too. @alimaredia can you please comment on this.

pritha-srivastava:
@alimaredia: I plan to use https://github.com/ceph/ceph/pull/35986/files as a reference in order to integrate these iam/user-policy tests in teuthology. Can you please confirm if something else needs to be done? I think an extra needs to be added for each iam test here in order to filter the tests out?

ravindrachoudhari:
@pritha-srivastava @alimaredia Added IAM policy tests section in the README file
Please review

README.rst
```diff
@@ -71,3 +71,45 @@ You can filter tests based on the attributes. There is a attribute named ``test_
 For running ``webidentity_test`` you'll need have Keycloak running.
 
 In order to run any STS test you'll need to add "iam" section to the config file. For further reference on how your config file should look check ``s3tests.conf.SAMPLE``.
+
+========================
+ IAM policy tests
+========================
+
+This is a set of IAM policy tests. Tests use the Boto2 and Boto3 libraries.
```
pritha-srivastava, line 79: I think these tests use only boto3, like you have mentioned at the end. Otherwise the README looks fine.

ravindrachoudhari: Yes, I will correct that.
Thank you

pritha-srivastava:
Ceph changes for vstart and teuthology here: https://github.com/ceph/ceph/pull/46660

pritha-srivastava:
@ravindrachoudhari : can you please add this attr - 'test_of_iam' to each test, in order to filter these tests in teuthology?

README.rst
pritha-srivastava: looks fine now.

ravindrachoudhari:
> @ravindrachoudhari : can you please add this attr - 'test_of_iam' to each test, in order to filter these tests in teuthology?

Sure, I will add @attr('test_of_iam') for all tests

ravindrachoudhari:
> > @ravindrachoudhari : can you please add this attr - 'test_of_iam' to each test, in order to filter these tests in teuthology?
> 
> Sure, I will add @attr('test_of_iam') for all tests

@pritha-srivastava Should I remove @attr('user-policy') or should I keep both @attr('user-policy') and @attr('test_of_iam')

pritha-srivastava:
> > > @ravindrachoudhari : can you please add this attr - 'test_of_iam' to each test, in order to filter these tests in teuthology?
> > 
> > 
> > Sure, I will add @attr('test_of_iam') for all tests
> 
> @pritha-srivastava Should I remove @attr('user-policy') or should I keep both @attr('user-policy') and @attr('test_of_iam')

you can keep both

ravindrachoudhari:
> > > > @ravindrachoudhari : can you please add this attr - 'test_of_iam' to each test, in order to filter these tests in teuthology?
> > > 
> > > 
> > > Sure, I will add @attr('test_of_iam') for all tests
> > 
> > 
> > @pritha-srivastava Should I remove @attr('user-policy') or should I keep both @attr('user-policy') and @attr('test_of_iam')
> 
> you can keep both

Done

ravindrachoudhari:
@pritha-srivastava did you get chance to take the teuthology run?

README.rst
```diff
@@ -71,3 +71,45 @@ You can filter tests based on the attributes. There is a attribute named ``test_
...
+and then run::
+
+	./bootstrap
+
+You will need to create a configuration file with the location of the
+service and two different credentials. A sample configuration file named
+``s3tests.conf.SAMPLE`` has been provided in this repo. This file can be
+used to run the iam policy tests on a Ceph cluster started with vstart.
+
+Note: These iam policy tests uses two users, "iam" and "s3 alt" as mentioned in s3tests.conf.SAMPLE.
```
alimaredia, line 95: @ravindrachoudhari A lot of the information below and above this section is redundant. Could you have this section contain the just following:
1. What the IAM policy tests cover
2. How to filter out IAM policy tests
3. What a radosgw user needs to be configured with to run these tests
4. Which users in the example config file are already configured to run these tests from a Ceph cluster configured with vstart.sh

alimaredia:
@pritha-srivastava let's not make any changes to vstart.sh

> @alimaredia: I plan to use https://github.com/ceph/ceph/pull/35986/files as a reference in order to integrate these iam/user-policy tests in teuthology. Can you please confirm if something else needs to be done? I think an extra needs to be added for each iam test here in order to filter the tests out when s3tests are executed?

@pritha-srivastava Yes an extra attr needs to be added, which you have seemed to discuss. Are you going to be adding in an `iam` directory in `ceph/qa/suites/rgw` that will run s3-tests to add the iam policy tests?

README.rst
ravindrachoudhari: Done. Please have a look.

alimaredia: Looks good to me!

ravindrachoudhari:
@pritha-srivastava @alimaredia can we merge this PR if everything is fine? or are we waiting for teuthology changes(https://github.com/ceph/ceph/pull/46660) to get merged first?

pritha-srivastava:
> @pritha-srivastava @alimaredia can we merge this PR if everything is fine? or are we waiting for teuthology changes([ceph/ceph#46660](https://github.com/ceph/ceph/pull/46660)) to get merged first?

@ravindrachoudhari : I remember we had discussed in the stand-up that https://github.com/ceph/ceph/pull/46660 needs to get merged first, and that if you are following the PR, I am working on it.

s3tests_boto3/functional/__init__.py
alimaredia: yes it is, could you remove it from this declaration.

alimaredia:
@ketanarlulkar @ravindrachoudhari The Ceph PR https://github.com/ceph/ceph/pull/46660 is ready to go. Could you squash your commits on this PR and then I'll merge this (and the Ceph PR) in conjunction.

ravindrachoudhari:
> @ketanarlulkar @ravindrachoudhari The Ceph PR [ceph/ceph#46660](https://github.com/ceph/ceph/pull/46660) is ready to go. Could you squash your commits on this PR and then I'll merge this (and the Ceph PR) in conjunction.

@alimaredia removed the region as suggested and done the commit's squash.

alimaredia:
@ravindrachoudhari great! I will double check that both this PR and the Ceph are ready to merge with @cbodley in case he doesn't know about them and merge both ASAP.

ravindrachoudhari:
> @ravindrachoudhari great! I will double check that both this PR and the Ceph are ready to merge with @cbodley in case he doesn't know about them and merge both ASAP.

Thank you.

soumyakoduri:
@pritha-srivastava @ravindrachoudhari @alimaredia ... the new tests added caused regression in dbstore test-suite - http://qa-proxy.ceph.com/teuthology/soumyakoduri-2022-06-29_09:47:08-rgw-wip-skoduri-dbstore-lc-distro-default-smithi/6905136/teuthology.log

Please include 'fails_on_dbstore' attr as well if you find any test cases failing against dbstore. Will submit new PR to fix this.

cbodley:
thanks @soumyakoduri, i opened https://tracker.ceph.com/issues/56418 to track this

soumyakoduri:
> thanks @soumyakoduri, i opened https://tracker.ceph.com/issues/56418 to track this

Pushed https://github.com/ceph/s3-tests/pull/460 as an interim workaround. But I think the right fix is to skip these iam/sts tests if iam {} section is not added to s3tests.conf instead of throwing error . Have raised https://tracker.ceph.com/issues/56421 to fix the test cases.

PR #455, opened 2022-05-19
dbstore: mark 'test_set_bucket_tagging' as 'fails_on_dbstore'

soumyakoduri:
As per below teuthology results, tests_set_bucket_tagging fails on
dbstore

http://qa-proxy.ceph.com/teuthology/soumyakoduri-2022-05-19_16:46:01-rgw:dbstore-wip-skoduri-dbstore-vstart-distro-basic-smithi/6841399/teuthology.log

Signed-off-by: Soumya Koduri <skoduri@redhat.com>


soumyakoduri:
This failure could  be because I was using older builds. Will check with latest build and reopen if issue still persists.

PR #456, opened 2022-05-20
Fix AttributeError

edta:
In eq statement status and error_code must be used not e.status and e.error_code


alimaredia:
duplicate of: https://github.com/ceph/s3-tests/pull/479

PR #476, opened 2022-11-22
pin botocore to resolve v2 signature failures

cbodley:
it looks like ubuntu 20.04 recently updated its botocore version, and introduced a regression in v2 signatures that causes 4 test failures: https://tracker.ceph.com/issues/58059

pin `botocore <1.28.0` in requirements.txt to avoid this regression


yuvalif (approved):

cbodley:
@alimaredia this regression showed up on pacific/quincy/main, so i've cherry-picked this for each

PR #479, opened 2022-11-30
fix `test_bucket_create_exists`

utkarsh-pro:
This PR attempts to fix `test_bucket_create_exists` by using appropriate status and error_code.

Signed-off-by: Utkarsh Srivastava <srivastavautkarsh8097@gmail.com>


TejasC88 (approved):

utkarsh-pro:
@TejasC88 can we merge this PR?

0xavi0:
Is this test passing even with the fixes? 

I found the same issue and tested the same code proposed in the PR, but the problem is that `rgw` from the main branch is not returning `BucketAlreadyOwnedByYou` (if my understanding is correct that error code is not even defined in: `rgw_common.cc`)

PR #482, opened 2023-01-22
replace deprecated nose with pytest

cbodley:
adds tox.ini to drive tests: https://tox.wiki/en/3.24.5/example/pytest.html
adds pytest fixtures for setup/teardown: https://docs.pytest.org/en/stable/how-to/fixtures.html
replace nose `@attr()`s with pytest markers: https://docs.pytest.org/en/stable/example/markers.html#mark-examples
replace nose `SkipTest` exception with `pytest.skip()`: https://docs.pytest.org/en/stable/how-to/skipping.html
replace nose assertions with python `assert`: https://docs.pytest.org/en/stable/how-to/assert.html


s3tests/functional/test_headers.py
```diff
@@ -178,28 +180,18 @@ def wrap(func):
 #
 
 @tag('auth_common')
```
alimaredia, line 182: @cbodley any idea if there's any use to this tag? If not can we go ahead and remove them in a follow up PR?

cbodley: good question, i left it there because i didn't understand it 🤷 

but looking through the git history, i see that it was originally added in https://github.com/ceph/s3-tests/pull/84, where it was used to filter between auth_common/auth_aws2/auth_aws4:
```
$ S3_USE_SIGV4=1 S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests/functional/test_headers.py -a auth_aws4
```
i'm not sure why the PR added a new `def tag(*tags)` function for this filtering instead of using nose's `@attr()` like we did for fails_on_rgw etc. `@tag` apparently worked with `nosetests -a auth_aws4`, but `pytest -m auth_aws4` doesn't:
```
$ pytest -m auth_aws4
collected 772 items / 772 deselected / 0 selected
```
so i'll change these to pytest markers the same way i did for fails_on_rgw and the others, and drop the `tag()` function

alimaredia: The only `tags` I see are `auth_common`, `auth_aws4`, and `auth_aws2`. If we're not using the tags for anything want to just remove them all together?

cbodley: i think it's useful to keep them for filtering. with that latest commit:
```
$ pytest -m auth_aws4
collected 772 items / 735 deselected / 37 selected
s3tests/functional/test_headers.py::test_object_create_bad_md5_invalid_garbage_aws4 ...
```

alimaredia (approved):
This looks grea! I might have some changes to add to the README but I'll do a follow up PR with that.

README.rst
```diff
@@ -22,37 +18,33 @@ used to run the s3 tests on a Ceph cluster started with vstart.
...
+	S3TEST_CONF=aws.conf tox -- -m 'not fails_on_aws'
 
 Most of the tests have both Boto3 and Boto2 versions. Tests written in
 Boto2 are in the ``s3tests`` directory. Tests written in Boto3 are
 located in the ``s3test_boto3`` directory.
 
 You can run only the boto3 tests with::
 
-        S3TEST_CONF=your.conf ./virtualenv/bin/nosetests -v -s -A 'not fails_on_rgw' s3tests_boto3.functional
+	S3TEST_CONF=your.conf tox -- -m 'not fails_on_aws' s3tests_boto3/functional
```
shirady, line 47: Hello :wave:
@cbodley @alimaredia 
Would you please explain why did you change it from 'not fails_on_rgw' to 'not fails_on_aws'?

cbodley: oops, thanks. that was probably a copy/paste mistake from the change above

cbodley: https://github.com/ceph/s3-tests/pull/514

nadavMiz: @cbodley can you explain what exactly this flag does? if boto3 and boto2 tests are in separate folders, why do we need an extra flag to run boto3 tests? shouldn't running the test in the boto3 folder be enough?

PR #487, opened 2023-02-01
test_post_object_upload_size_below_minimum_rgw_max_chunk_size: new testcase for size validation bug

robbat2:
If all of the following are true `ERR_TOO_SMALL` is wrongly returned:
- the bufferlist contains multiple items
- the last item has non-zero length
- that length is smaller than the POST Policy

The check should be `(ofs < min_len)` instead of `(len < min_len)`.

This is further confirmed by the fact that the next line does `s->obj_size = ofs`;

The bug was refactored in 2016, but was introduced in Oct 2012, when
this functionality was first added to RGW in commit 7bb3504d3f0974e9863f536e9af0ce8889d6888f.

Reference: https://github.com/ceph/ceph/blob/933a42f9af349b3b222270e7f19f1fe151d89e8e/src/rgw/rgw_op.cc#L4474-L4513
Reference: https://github.com/ceph/ceph/commit/7bb3504d3f0974e9863f536e9af0ce8889d6888f
Signed-off-by: Robin H. Johnson <rjohnson@digitalocean.com>

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -2818,6 +2818,53 @@ def test_post_object_upload_size_below_minimum():
     r = requests.post(url, files=payload, verify=get_config_ssl_verify())
     assert r.status_code == 400
 
+def test_post_object_upload_size_rgw_chunk_size_bug():
+    # Test for https://tracker.ceph.com/issues/58627
+    # TODO: if this value is different in Teuthology runs, this would need tuning
+    # https://github.com/ceph/ceph/blob/main/qa/suites/rgw/verify/striping%24/stripe-greater-than-chunk.yaml
+    _rgw_max_chunk_size = 4 * 2**20 # 4MiB
+    min_size = _rgw_max_chunk_size
+    max_size = _rgw_max_chunk_size * 3
+    # [(chunk),(small)]
+    test_payload_size = _rgw_max_chunk_size + 200 # extra bit to push it over the chunk boundary
+    # it should be valid when we run this test!
+    assert test_payload_size > min_size
+    assert test_payload_size < max_size
+
+    bucket_name = get_new_bucket()
+    client = get_client()
+
+    url = _get_post_url(bucket_name)
+    utc = pytz.utc
+    expires = datetime.datetime.now(utc) + datetime.timedelta(seconds=+6000)
+
+    policy_document = {"expiration": expires.strftime("%Y-%m-%dT%H:%M:%SZ"),\
+    "conditions": [\
+    {"bucket": bucket_name},\
+    ["starts-with", "$key", "foo"],\
+    {"acl": "private"},\
+    ["starts-with", "$Content-Type", "text/plain"],\
+    ["content-length-range", min_size, max_size],\
+    ]\
+    }
+
+    test_payload = 'x' * test_payload_size
+
+    json_policy_document = json.JSONEncoder().encode(policy_document)
+    bytes_json_policy_document = bytes(json_policy_document, 'utf-8')
+    policy = base64.b64encode(bytes_json_policy_document)
+    aws_secret_access_key = get_main_aws_secret_key()
+    aws_access_key_id = get_main_aws_access_key()
+
+    signature = base64.b64encode(hmac.new(bytes(aws_secret_access_key, 'utf-8'), policy, hashlib.sha1).digest())
+
+    payload = OrderedDict([ ("key" , "foo.txt"),("AWSAccessKeyId" , aws_access_key_id),\
+    ("acl" , "private"),("signature" , signature),("policy" , policy),\
+    ("Content-Type" , "text/plain"),('file', (test_payload))])
+
+    r = requests.post(url, files=payload, verify=get_config_ssl_verify())
+    assert r.status_code == 204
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/test_s3.py
```diff
@@ -2817,6 +2817,53 @@ def test_post_object_upload_size_below_minimum():
...
+def test_post_object_upload_size_rgw_chunk_size_bug():
+    # TODO: if this value is different in Teuthology runs, this would need tuning
+    _rgw_max_chunk_size = 4 * 2**20 # 4MiB
+    min_size = _rgw_max_chunk_size * 2
+    max_size = _rgw_max_chunk_size * 3
+    # [(chunk),(small)]
+    test_payload_size = min_size * 2 + 200 # extra bit to push it over the chunk boundary
+    # it should be valid when we run this test!
+    assert test_payload_size > min_size
+    assert test_payload_size < max_size
```
cbodley, line 2831: double-checking my math here..
```
rgw_max_chunk_size = 4M
min_size = 8M
max_size = 12M
test_payload_size = 16M + 200
```
so won't this `max_size` assertion fail?

in the interest of a minimal reproducer, could we just use `test_payload_size = _rgw_max_chunk_size + 200`?

s3tests_boto3/functional/test_s3.py
```diff
@@ -2817,6 +2817,53 @@ def test_post_object_upload_size_below_minimum():
 
     r = requests.post(url, files=payload, verify=get_config_ssl_verify())
     assert r.status_code == 400
+    assert r.error_code == 'EntityTooSmall'
+
+def test_post_object_upload_size_rgw_chunk_size_bug():
+    # TODO: if this value is different in Teuthology runs, this would need tuning
```
cbodley, line 2823: the rgw suite does vary the *stripe* size, but the chunk size stays constant at 4M: https://github.com/ceph/ceph/blob/main/qa/suites/rgw/verify/striping%24/stripe-greater-than-chunk.yaml

s3tests_boto3/functional/test_s3.py
```diff
@@ -2817,6 +2817,53 @@ def test_post_object_upload_size_below_minimum():
...
+    _rgw_max_chunk_size = 4 * 2**20 # 4MiB
+    min_size = _rgw_max_chunk_size * 2
+    max_size = _rgw_max_chunk_size * 3
```
cbodley, line 2826: we try to keep rgw specifics out s3-tests, but i'm happy with the way you've done this 👍 it exercises a corner case for rgw, but should succeed regardless of the server. since it mentions an rgw bug, please add a comment that links to https://tracker.ceph.com/issues/58627

s3tests_boto3/functional/test_s3.py
robbat2: oops, editing fail when I cleaned this from internal version

robbat2: Fixed

s3tests_boto3/functional/test_s3.py
robbat2: Done

s3tests_boto3/functional/test_s3.py
robbat2: Thanks, linked

s3tests_boto3/functional/test_s3.py
```diff
@@ -2817,6 +2817,55 @@ def test_post_object_upload_size_below_minimum():
...
+
+    signature = base64.b64encode(hmac.new(bytes(aws_secret_access_key, 'utf-8'), policy, hashlib.sha1).digest())
+
+    payload = OrderedDict([ ("key" , "foo.txt"),("AWSAccessKeyId" , aws_access_key_id),\
+    ("acl" , "private"),("signature" , signature),("policy" , policy),\
+    ("Content-Type" , "text/plain"),('file', (test_payload))])
+
+    r = requests.post(url, files=payload, verify=get_config_ssl_verify())
+    assert r.status_code == 400
+    assert r.error_code == 'EntityTooSmall'
```
cbodley, line 2868: shouldn't we expect success in this case?

cbodley: ping @robbat2

robbat2: Good catch, can you please amend and merge? DM me for $reasons.

cbodley: sure. no explanation necessary

cbodley (approved):

s3tests_boto3/functional/test_s3.py
```diff
@@ -2817,6 +2817,54 @@ def test_post_object_upload_size_below_minimum():
 
     r = requests.post(url, files=payload, verify=get_config_ssl_verify())
     assert r.status_code == 400
+    assert r.error_code == 'EntityTooSmall'
```
cbodley, line 2820: this fails with `AttributeError: 'Response' object has no attribute 'error_code'`

this is a call to `requests.post()` which doesn't understand these s3 error codes. i think we'd have to parse the xml response to get at it. i'm going to remove this commit for now

Issue #493, opened 2023-02-19
est_s3.py::test_bucket_list_distinct fails 


vardhanv:
`
s3tests_boto3/functional/__init__.py:315: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
s3tests_boto3/functional/__init__.py:285: in teardown
    nuke_prefixed_buckets(prefix=prefix)
s3tests_boto3/functional/__init__.py:159: in nuke_prefixed_buckets
    raise err
s3tests_boto3/functional/__init__.py:150: in nuke_prefixed_buckets
    nuke_bucket(client, bucket_name)
s3tests_boto3/functional/__init__.py:104: in nuke_bucket
    delete = client.delete_objects(Bucket=bucket,
.tox/py/lib/python3.11/site-packages/botocore/client.py:514: in _api_call
    return self._make_api_call(operation_name, kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <botocore.client.S3 object at 0x1033a4fd0>, operation_name = 'DeleteObjects'
api_params = {'Bucket': 'vishnu-ceph-z9cpnnzn83s0deb40-1', 'BypassGovernanceRetention': True, 'Delete': {'Objects': [{'Key': 'asdf', 'VersionId': 'null'}], 'Quiet': True}}

<snip>

E           botocore.exceptions.ClientError: An error occurred (InvalidArgument) when calling the DeleteObjects operation: x-amz-bypass-governance-retention is only applicable to Object Lock enabled buckets.

.tox/py/lib/python3.11/site-packages/botocore/client.py:938: ClientError

`

PR #494, opened 2023-02-19
Fix for #493

vardhanv:
Includes changes from pull - https://github.com/ceph/s3-tests/pull/490, as they are a dependency


vardhanv:
@cbodley could you advise on who should review this?

cbodley:
> Includes changes from pull - #490, as they are a dependency Fixes: #493

would you please rebase now that #490 merged?

PR #500, opened 2023-03-07
lifecycle: Fix test_lifecycle_expiration_header_* testcases

soumyakoduri:
Few checks were incorrectly mapped when switched to 'assert' (quincy -https://github.com/ceph/s3-tests/blob/ceph-quincy/s3tests_boto3/functional/test_s3.py#L9886). 

This commit fixes the same.


cbodley (approved):
sorry again!

cbodley:
cherry-picked to ceph-reef and ceph-master, thanks!

PR #502, opened 2023-03-12
s3website: collections.Container removed from python3.10

cbodley:
this was added to [collections.abc](https://docs.python.org/3/library/collections.abc.html) in python 3.3, and removed from collections in 3.10

> AttributeError: module 'collections' has no attribute 'Container'

Fixes: https://tracker.ceph.com/issues/58960

### DIFF
--- a/s3tests/functional/test_s3_website.py
+++ b/s3tests/functional/test_s3_website.py
@@ -1,5 +1,5 @@
 import sys
-import collections
+from collections.abc import Container
 import pytest
 import string
 import random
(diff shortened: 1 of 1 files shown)


alimaredia (approved):

alimaredia:
@cbodley lgtm. Do you have a run of s3-tests that shows this passes?

cbodley:
> @cbodley lgtm. Do you have a run of s3-tests that shows this passes?

not yet. i can batch it together with https://github.com/ceph/s3-tests/pull/503 when i'm ready to test https://github.com/ceph/ceph/pull/50504

cbodley:
@alimaredia cherry-picked to ceph-master and ceph-reef

PR #503, opened 2023-03-13
boto3: multi-object-delete tests use client.delete_objects()

cbodley:
when the tests were converted from boto2, they were rewritten as loops over `client.delete_object()`. switch back to multi-delete

the `test_versioning_multi_object_delete_with_marker_create` test case now checks the `delete_objects()` response, and uncovered the rgw bug https://tracker.ceph.com/issues/59048


cbodley:
passed qa in https://pulpito.ceph.com/cbodley-2023-03-14_21:22:40-rgw-wip-cbodley-testing-distro-default-smithi/

cbodley:
@alimaredia i think it's been long enough since the merge of https://github.com/ceph/ceph/pull/50504; find the successful QA results above

alimaredia (approved):

PR #505, opened 2023-03-30
test_sse_s3_default_multipart_upload verifies encryption header

cbodley:
there's a `test_sse_s3_default_multipart_upload` test case, but it wasn't verifying that the upload was actually encrypted

for https://github.com/ceph/ceph/pull/49409 https://tracker.ceph.com/issues/59218


alimaredia (approved):

alimaredia:
@cbodley besides backporting `ceph-master` should I wait for this patch to be backported to `ceph-quincy` and `ceph-reef` once those two Ceph backports are complete? I see https://tracker.ceph.com/issues/59355 and https://tracker.ceph.com/issues/59356 on pending backport atm.

PR #506, opened 2023-04-05
Json s3tests

galsalomon66:
- adding tests per JSON input types, which includes various SQL operators. 
- alignment of some of the test routines upon removing the XML tags(\<Payload\>\<Records\>\<Payload\>) from the s3select results.

### DIFF
--- a/s3tests_boto3/functional/test_s3select.py
+++ b/s3tests_boto3/functional/test_s3select.py
@@ -75,7 +75,7 @@ def generate_s3select_expression_projection(bucket_name,obj_name):
         epsilon = float(0.00001) 
 
         # both results should be close (epsilon)
-        assert(  abs(float(res.split("\n")[1]) - eval(e)) < epsilon )
+        assert(  abs(float(res.split("\n")[0]) - eval(e)) < epsilon )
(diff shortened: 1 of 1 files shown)


cbodley:
@galsalomon66 is this ready to merge?

galsalomon66:
> @galsalomon66 is this ready to merge?

yes (tuethology ran with that)

cbodley (approved):

galsalomon66:
@cbodley 
how to backport this PR (into reef)?

cbodley:
similarly to how you cherry-picked to ceph-master, we'd cherry-pick to the ceph-reef branch. just be careful with the timing, so we don't introduce failures in other reef tests

galsalomon66:
> similarly to how you cherry-picked to ceph-master, we'd cherry-pick to the ceph-reef branch. just be careful with the timing, so we don't introduce failures in other reef tests

ok
1. cherry-picked into ceph-reef (same as ceph-master). so  ...  i understand no tracker? no scripts for backport?
2. as for timing .... i guess both PR's should merge into reef at the same time.

PR #266, opened 2019-03-26
add multipart ranges to encryption tests

theanalyst:

PR #516, opened 2023-06-21
test_encryption_sse_c_unaligned_multipart_upload fails_on_dbstore

cbodley:
fixes a new dbstore failure introduced by https://github.com/ceph/s3-tests/pull/266. `DBBucket::update_container_stats()` is a noop, so the response headers `x-rgw-object-count` and `x-rgw-bytes-used` are not initialized

example: http://qa-proxy.ceph.com/teuthology/soumyakoduri-2023-06-20_12:37:42-rgw-wip-rgw-yield-work-distro-default-smithi/7309055/teuthology.log


cbodley:
i'm actually going to remove most references to these x-rgw- response headers, and use `list_objects()` instead when checking object counts and sizes

Issue #315, opened 2019-10-27
Ceph Tests make specific assumptions about RGW custom headers


romayalon:
The following list of tests make assumptions about custom RGW headers in the S3 responses :
1. s3tests_boto3.functional.test_s3.test_abort_multipart_upload
2. s3tests_boto3.functional.test_s3.test_bucket_head_extended
3. s3tests_boto3.functional.test_s3.test_encryption_sse_c_multipart_upload
4. s3tests_boto3.functional.test_s3.test_multipart_upload 
5. s3tests_boto3.functional.test_s3.test_sse_kms_multipart_upload

The tests expect a response that contain custom headers like  ```x-rgw-bytes-used``` and ```x-rgw-object-count```, etc. which are not a part of the Amazon AWS  S3 spec.

PR #517, opened 2023-06-21
remove reliance on x-rgw-object-count and x-rgw-bytes-used headers

cbodley:
stop using `head_bucket()` to fetch these response headers, and use `list_objects_v2()` instead to count objects and sizes. this allows us to remove several `fails_on_aws` and `fails_on_dbstore` markers

Fixes: #315

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -5932,7 +5932,6 @@ def _check_content_using_range(key, bucket_name, data, step):
         body = _get_body(response)
         assert body == data[ofs:end+1]
 
-@pytest.mark.fails_on_aws
 @pytest.mark.fails_on_dbstore
 def test_multipart_upload():
     bucket_name = get_new_bucket()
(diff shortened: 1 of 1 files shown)


alimaredia (approved):

alimaredia:
@cbodley this looks good to me. Can I merge this and then run the master branch on the dbstore and verify (with limit 1) rgw suites?

cbodley:
@alimaredia tests are queued in https://pulpito.ceph.com/cbodley-2023-06-21_18:30:33-rgw-main-distro-default-smithi/, let's wait on that

soumyakoduri (approved):

cbodley:
some of these test cases failed in the rgw/dbstore suite, so i had to restore some of the `fails_on_dbstore` tags; those failures weren't only due to the x-rgw- headers

alimaredia:
got it, when you post a clean run I'll go ahead and merge the PR

soumyakoduri:
cherry-picked to `ceph-master`

PR #522, opened 2023-06-29
Remove filtering from boto3 test example for simplicity

nadavMiz:
having the filter from previous examples might be confusing because It can be interpreted as a needed part to run only boto3 tests. I removed it to make the example as simple as possible

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,7 +22,7 @@ Once you have that file copied and edited, you can run the tests with::
 
 You can specify which directory of tests to run::
 
-	S3TEST_CONF=your.conf tox s3tests_boto3/functional
+	S3TEST_CONF=your.conf tox -- s3tests_boto3/functional
 
 You can specify which file of tests to run::
 
@@ -44,7 +44,7 @@ located in the ``s3test_boto3`` directory.
 
 You can run only the boto3 tests with::
 
-	S3TEST_CONF=your.conf tox -- -m 'not fails_on_rgw' s3tests_boto3/functional
+	S3TEST_CONF=your.conf tox -- s3tests_boto3/functional
 
 ========================
  STS compatibility tests


README.rst
```diff
@@ -44,7 +44,7 @@ located in the ``s3test_boto3`` directory.
 
 You can run only the boto3 tests with::
 
-	S3TEST_CONF=your.conf tox -- -m 'not fails_on_rgw' s3tests_boto3/functional
+	S3TEST_CONF=your.conf tox s3tests_boto3/functional
```
cbodley, line 47: are you sure it doesn't still need the `--`? the path is for the `pytest` command line, not `tox`

nadavMiz: at least in my environment, it works either way. I think it's better without the `--` because it's consistent with the section about running a directory of tests:
https://github.com/ceph/s3-tests/blob/master/README.rst?plain=1#L23-L26

cbodley:
```
~/s3-tests $ S3TEST_CONF=s3tests.conf.SAMPLE tox s3tests_boto3/functional
usage: tox [-h] [--colored {yes,no}] [-v | -q] [--exit-and-dump-after seconds] [-c file] [--workdir dir] [--root dir] [--runner {virtualenv}] [--version] [--no-provision [REQ_JSON]] [--no-recreate-provision] [-r] [-x OVERRIDE]
           {run,r,run-parallel,p,depends,de,list,l,devenv,d,config,c,quickstart,q,exec,e,legacy,le} ...
tox: error: unrecognized arguments: s3tests_boto3/functional
hint: if you tried to pass arguments to a command use -- to separate them from tox ones
```

nadavMiz: changed it, maybe we should change https://github.com/ceph/s3-tests/blob/master/README.rst?plain=1#L23-L26 as well since it's the same command? what do you think?

cbodley: yes please!

nadavMiz: done

cbodley (approved):
thanks!

PR #527, opened 2023-09-28
add test_post_object_wrong_bucket() for CVE-2023-43040

cbodley:
test case for https://tracker.ceph.com/issues/63004

based on the [s3-tests.patch](https://tracker.ceph.com/attachments/download/6661/s3-tests.patch) provided with the bug report, but with the `bucket` field added to the payload (thanks Nicolas Lindae)


cbodley:
@alimaredia this was used to validate https://github.com/ceph/ceph/pull/53714, and will be needed on other branches to validate the backports

alimaredia (approved):

PR #529, opened 2023-11-02
sse: add config option and test case for default encryption

cbodley:
if default encryption is enabled globally, test that uploads get encrypted even if not requested by the upload of bucket encryption policy

relates to https://github.com/ceph/ceph/pull/52697


cbodley:
closing in favor of https://github.com/ceph/s3-tests/pull/530

PR #530, opened 2023-11-02
sse: add test for default forced sse s3 encryption

clwluvw:
Related to #52697


s3tests_boto3/functional/test_s3.py
```diff
@@ -12703,6 +12703,57 @@ def test_delete_bucket_encryption_kms():
...
+    response = client.put_object(Bucket=bucket_name, Key='testobj', Body=data)
+    assert response['ResponseMetadata']['HTTPHeaders']['x-amz-server-side-encryption'] == 'AES256'
+
+    response = client.get_object(Bucket=bucket_name, Key='testobj')
+    assert response['ResponseMetadata']['HTTPHeaders']['x-amz-server-side-encryption'] == 'AES256'
+    body = _get_body(response)
+    assert body == data
+
+@pytest.mark.encryption
+@pytest.mark.forced_sse_s3
```
cbodley, line 12734: using a `forced_sse_s3` tag instead of config option is fine :+1:

please also add the `sse_s3` tag so existing use of `!sse_s3` filters these out too

clwluvw: Done.

PR #532, opened 2023-11-15
object lock: test very large RetainUntilDate

cbodley:
https://tracker.ceph.com/issues/63537 reported that large dates (with year after 2107) got truncated when written. test with a later date, and check that get_object_retention() gives back the date we put


cbodley:
ping @alimaredia

alimaredia (approved):

cbodley:
thanks @alimaredia, cherry-picked to ceph-master

PR #537, opened 2023-12-19
iam: add tests for account-based IAM apis

cbodley:
renames the `test_of_iam` label to `iam_tenant` for the existing tenant-based iam tests

adds new `[iam root]` and `[iam alt root]` config sections for account root users, and a new set of `iam_account` and `iam_cross_account` test cases that use them

adds pytest fixtures `iam_root` and `iam_alt_root` that return the corresponding boto client, with cleanup logic to nuke any created users/roles/etc. all iam resources created by the tests have names starting with `iam name prefix = s3-tests-` and paths starting with `iam path prefix = /s3-tests/` so this cleanup logic doesn't nuke anything else in the aws account

adds account-based test cases for the following iam api categories:
* User
* AccessKey
* UserPolicy
* Group
* GroupPolicy
* Role
* RolePolicy
* OpenIDConnectProvider

verified to pass against aws when an account root user's credentials are provided in the [iam] section of s3tests.conf (haven't tested iam_cross_account tests yet)

TODO:
- [x] assuming another account's role uses same-account rules to access that account's resources
- [x] assuming another account's role uses cross-account rules to access the original user's resources
- [x] buckets/objects created by the assumed role belong to the role's account
- [x] test that identity policy can grant non-root users access to iam account apis

### DIFF
--- a/s3tests_boto3/functional/test_iam.py
+++ b/s3tests_boto3/functional/test_iam.py
@@ -16,6 +16,7 @@ from . import (
     make_iam_name,
     get_iam_path_prefix,
     get_new_bucket,
+    get_new_bucket_name,
     get_iam_s3client,
     get_alt_iam_client,
     get_alt_user_id,
@@ -878,10 +879,23 @@ def nuke_user_keys(client, name):
             except:
                 pass
 
+def nuke_user_policies(client, name):
+    p = client.get_paginator('list_user_policies')
+    for response in p.paginate(UserName=name):
+        for policy in response['PolicyNames']:
+            try:
+                client.delete_user_policy(UserName=name, PolicyName=policy)
+            except:
+                pass
+
 def nuke_user(client, name):
     # delete access keys, user policies, etc
     try:
         nuke_user_keys(client, name)
+    except:
+        pass
+    try:
+        nuke_user_policies(client, name)
     except:
         pass
     client.delete_user(UserName=name)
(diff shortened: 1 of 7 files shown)


s3tests_boto3/functional/test_iam.py
```diff
@@ -862,3 +866,219 @@ def test_verify_allow_iam_actions():
...
+        for user in response['Users']:
+            try:
+                nuke_user(client, user['UserName'])
+            except:
+                pass
+
+# fixture for iam account root user
+@pytest.fixture
+def iam_root(configfile):
+    client = get_iam_root_client()
```
pritha-srivastava, line 1035: @cbodley : just curious - how do we differentiate between a root user and other IAM users in rgw?

s3tests_boto3/functional/test_iam.py
```diff
@@ -1276,3 +1277,59 @@ def test_account_user_access_key_list(iam_root):
...
+        status, error_code = _get_status_and_error_code(e.response)
+        assert status == 403
+        assert error_code == 'AccessDenied'
+
+        # add a bucket policy that allows s3:ListBucket for the iam user's arn
+        policy = json.dumps({
+            'Version': '2012-10-17',
+            'Statement': [{
+                'Effect': 'Allow',
+                'Principal': {'AWS': user_arn},
```
pritha-srivastava, line 1462: is this also testing cross-account access?

s3tests_boto3/functional/test_iam.py
```diff
@@ -1718,3 +1719,122 @@ def test_account_role_update(iam_root):
...
+@pytest.mark.iam_account
+@pytest.mark.iam_role
+@pytest.mark.role_policy
+def test_account_role_policy_allow(iam_root):
+    path = get_iam_path_prefix()
+    user_name = make_iam_name('MyUser')
+    role_name = make_iam_name('MyRole')
+    session_name = 'MySession'
+
+    user = iam_root.create_user(UserName=user_name, Path=path)['User']
```
pritha-srivastava, line 2253: the user and the role, both belong to the same account, am I correct? If so we could modify the test case to test assume role for cross account access.

s3tests_boto3/functional/test_iam.py
```diff
@@ -1333,3 +1347,98 @@ def test_account_user_bucket_policy_allow(iam_root):
...
+            'Action': 's3:*',
+            'Resource': '*'
+            }]
+        })
+    policy_name = 'AllowStar'
+    iam_root.put_user_policy(UserName=name, PolicyName=policy_name, PolicyDocument=policy)
+
+    # the policy may take a bit to start working. retry until it returns
+    # something other than AccessDenied
+    retry_on('AccessDenied', 10, client.list_buckets)
```
pritha-srivastava, line 1624: there are other tests that we could add here, like let the root user attach a policy to an iam user which allows iam actions like 'PutUserPolicy' etc, and then test whether the iam user is able to perform those iam actions, but I am not sure whether an end user using rgw will be performing such actions and whether it is worth the effort to test this.

s3tests_boto3/functional/test_iam.py
```diff
@@ -1780,6 +1793,63 @@ def test_account_role_policy(iam_root):
     with pytest.raises(iam_root.exceptions.NoSuchEntityException):
         iam_root.delete_role_policy(RoleName=role_name, PolicyName=policy_name)
 
+@pytest.mark.role_policy
+@pytest.mark.iam_account
+def test_account_role_policy_managed(iam_root):
+    path = get_iam_path_prefix()
+    name = make_iam_name('name')
+    policy1 = 'arn:aws:iam::aws:policy/AmazonS3FullAccess'
```
pritha-srivastava, line 2192: aren't these managed policies supposed to exist somewhere?

s3tests_boto3/functional/test_iam.py
```diff
@@ -1908,3 +1912,377 @@ def test_account_role_policy_allow(iam_root):
...
+                'Resource': f'arn:aws:s3:::{bucket}'
+                }]
+            }))
+
+        # the access key may take a bit to start working. retry until it returns
+        # something other than InvalidAccessKeyId
+        e = assert_raises(ClientError, retry_on, 'InvalidAccessKeyId', 10, alts3.list_objects, Bucket=bucket)
+        status, error_code = _get_status_and_error_code(e.response)
+        assert status == 403
+        assert error_code == 'AccessDenied'
```
pritha-srivastava, line 2421: In my understanding, if the bucket policy allows the s3 action, then it must be eventually allowed.

s3tests_boto3/functional/test_iam.py
```diff
@@ -1908,3 +1912,377 @@ def test_account_role_policy_allow(iam_root):
...
+
+        # add a user policy that allows s3 actions
+        alt_root.put_user_policy(UserName=alt_name, PolicyName='AllowStar', PolicyDocument=json.dumps({
```
pritha-srivastava, line 2424: this policy will allow all s3 actions only in the current account.

s3tests_boto3/functional/test_iam.py
```diff
@@ -1908,3 +1912,377 @@ def test_account_role_policy_allow(iam_root):
...
+def test_cross_account_root_bucket_acl_grant_canonical_id(iam_root, iam_alt_root):
+    roots3 = get_iam_root_client(service_name='s3')
+    alts3 = get_iam_alt_root_client(service_name='s3')
+    grantee = 'id=' + get_iam_alt_root_user_id()
+    _test_cross_account_root_bucket_acl(roots3, alts3, grantee)
+
+@pytest.mark.iam_account
+@pytest.mark.iam_cross_account
+def test_cross_account_root_bucket_acl_grant_account_email(iam_root, iam_alt_root):
+    roots3 = get_iam_root_client(service_name='s3')
```
pritha-srivastava, line 2706: cross account access is mainly provided by bucket policy and role policy (and not by user policy, iiuc), so we should cover some cases uses role arns and role session arns along with bucket policies. (https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies-cross-account-resource-access.html)

s3tests_boto3/functional/test_iam.py
cbodley: in https://github.com/ceph/ceph/pull/54333/commits/b90e93c9543721c4b3e6d870ff4754df2cf59fbf i extended enum RGWIdentityType with a TYPE_ROOT that's used when a user is created with the --account-root option
```c++
if (s->identity->get_identity_type() == TYPE_ROOT) { // account root user
```

s3tests_boto3/functional/test_iam.py
cbodley: there are a ton of 'aws managed policies' listed in https://docs.aws.amazon.com/aws-managed-policy/latest/reference/policy-list.html that exist in aws

i chose a small number of those to hard-code into rgw in https://github.com/ceph/ceph/pull/54333/commits/75d295221b203620a1d169c85ae46ff5857dda8e

basically just the FullAccess and ReadOnlyAccess policies for S3, SNS, STS, IAM

s3tests_boto3/functional/test_iam.py
cbodley: i agree that there are more interesting things to test here:
* assuming another account's role uses same-account rules to access that account's resources
* assuming another account's role uses cross-account rules to access the original user's resources
* buckets/objects created by the assumed role belong to the role's account

but i still think this test case is useful to verify that the role policy is evaluated in the first place. each of the UserPolicy/GroupPolicy/RolePolicy sections have a test case in this form

s3tests_boto3/functional/test_iam.py
cbodley: > or is it that unless an iam user has permissions to perform an action in its own account, it won't be allowed to perform those s3 actions even if a bucket policy allows it in a different account?

that's right. "Cross-account policy evaluation logic" is described in https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic-cross-account.html. that was very simple to implement in https://github.com/ceph/ceph/pull/54333/commits/375445d64eeca324ef81c3dc8dcfedf9a98cd720

this cross-account policy evaluation is one of the major differences between 'account users' and normal 'tenant users' in rgw. while an account user starts with no permissions at all, tenant users do have permissions by default (a new user can create buckets and upload objects)

as a result, tenant users always use the same-account policy evaluation rules you linked above. i hadn't realized that before, but it gives our existing UserPolicy and RolePolicy APIs the ability to grant access to the resources of other users in that tenant

s3tests_boto3/functional/test_iam.py
cbodley: these are testing the cross-account policy evaluation rules, where any identity policy works as long as there's also a resource policy. i stuck with user policy in these tests because it's the easiest identity policy to apply (you don't have to create a group or role also)

AssumeRole helps with cross-access access, but in a different way - it turns cross-account policy evaluation into same-account evaluation. the iam_tenant tests should already cover that

session policy is evaluated differently from identity policy, but we can test those differences in the context of same-account policy evaluation. the iam_tenant tests should cover that too

cbodley:
@pritha-srivastava thanks for the review! i updated the pr description with a TODO list of test cases to add

s3tests_boto3/functional/test_iam.py
pritha-srivastava: ok, how about `Customer Managed Policies`, do we have any plan to support that?

cbodley: > ok, how about `Customer Managed Policies`, do we have any plan to support that?

i hadn't planned to here. we'd need an efficient way to store them somewhere in rados and make them available to each request that uses them

i added these aws-managed ones because iam users/roles start with no policy, and the FullAccess and ReadOnlyAccess ones are a really convenient way to add blanket identity policies. they're also "free" to implement since we don't have to load the policy from rados

cbodley:
hey @alimaredia, i'm close to getting account stuff passing in teuthology. just wanted to get this on your radar

alimaredia:
@cbodley are these still underdevelopment? I see you force pushed just a couple hours ago and I don't see a run linked yet. Have these only been run locally?

cbodley:
@alimaredia this is ready for review. i'm still iterating on https://github.com/ceph/ceph/pull/54333 to get a clean run. all tests pass locally but i'm working through differences in configuration between rgw/verify rgw/sts etc

cbodley:
ping @alimaredia 

https://pulpito.ceph.com/cbodley-2024-03-13_12:14:22-rgw-wip-rgw-account-v3-distro-default-smithi/ has a clean run against this s3-tests branch

alimaredia (approved):

cbodley:
pulled in sns commits from https://github.com/cbodley/s3-tests/pull/1

accounts pr passed qa against this in https://pulpito.ceph.com/cbodley-2024-04-12_12:44:47-rgw-wip-rgw-account-v3-distro-default-smithi/

cbodley:
haven't cherry-picked for master or squid yet. for squid, i still have one batch pending qa. for main, i'll wait til later in the week

cbodley:
cherry-picked for ceph-squid

cbodley:
cherry-picked for ceph-master too

PR #542, opened 2024-02-04
Test CORS for presigned put_object

tobias-urdin:
This adds CORS testing for presigned put_object
URLs.


tobias-urdin:
superceeded by https://github.com/ceph/s3-tests/pull/544

PR #544, opened 2024-02-05
Add more testing for presigned URLs

tobias-urdin:
This improves the testing for presigned URLs for
both get_object and put_object when using
generate_presigned_url().

It covers the case where you pass for example
a x-amz-acl (ACL in params for generated_presigned_url) header that should be signed.

Tests the regression in [1].

[1] https://tracker.ceph.com/issues/64308

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -3522,6 +3522,36 @@ def test_object_put_authenticated():
     response = client.put_object(Bucket=bucket_name, Key='foo', Body='foo')
     assert response['ResponseMetadata']['HTTPStatusCode'] == 200
 
+def _test_object_presigned_put_object_with_acl(client=None):
+    if client is None:
+        client = get_client()
+
+    bucket_name = get_new_bucket(client)
+    key = 'foo'
+
+    params = {'Bucket': bucket_name, 'Key': key, 'ACL': 'private'}
+    url = client.generate_presigned_url(ClientMethod='put_object', Params=params, HttpMethod='PUT')
+
+    data = b'hello world'
+    headers = {'x-amz-acl': 'private'}
+    res = requests.put(url, data=data, headers=headers, verify=get_config_ssl_verify())
+    assert res.status_code == 200
+
+    params = {'Bucket': bucket_name, 'Key': key}
+    url = client.generate_presigned_url(ClientMethod='get_object', Params=params, HttpMethod='GET')
+
+    res = requests.get(url, verify=get_config_ssl_verify())
+    assert res.status_code == 200
+    assert res.text == 'hello world'
+
+def test_object_presigned_put_object_with_acl():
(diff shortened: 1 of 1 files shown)


tobias-urdin:
```
S3TEST_CONF=s3tests.conf.SAMPLE tox -- s3tests_boto3/functional/test_s3.py -k 'test_presigned_post or test_object_presigned_put_object_with_acl or test_cors_presigned or test_object_raw_get_x_amz_expires_not_expired' -v
```

cbodley (approved):
>tox -- s3tests_boto3/functional/test_s3.py -k 'test_object_presigned_put_object_with_acl or test_cors_presigned or test_object_raw_get_x_amz_expires_not_expired' -v

i managed to run these against aws and saw them pass against both us-east-1 and -2 👍 

that took a few minor changes to work around errors like AclWithBlockPublicAccessError and IllegalLocationConstraintException, and it didn't exercise tenants. i used the same iam user credentials for [s3 main] [s3 alt] and [s3 tenant]

s3tests_boto3/functional/test_s3.py
```diff
@@ -3522,6 +3522,37 @@ def test_object_put_authenticated():
     response = client.put_object(Bucket=bucket_name, Key='foo', Body='foo')
     assert response['ResponseMetadata']['HTTPStatusCode'] == 200
 
+def _test_object_presigned_put_object_with_acl(client=None):
+    if client is None:
+        client = get_client()
+
+    bucket_name = get_new_bucket()
+    client.create_bucket(Bucket=bucket_name)
```
cbodley, line 3530: [get_new_bucket()](https://github.com/ceph/s3-tests/blob/da91ad8bbf899c72199df35b69e9393c706aabee/s3tests_boto3/functional/__init__.py#L584-L596) actually creates the bucket using the default `get_client()`, so `client.create_bucket()` always recreates it. in the tenant test, this recreation is expected to fail:

> An error occurred (BucketAlreadyExists) when calling the CreateBucket operation: Cannot modify existing access control policy

strangely, i'm only seeing it fail that way in the rgw/dbstore suite

you can either call `get_new_bucket(client)` with the correct client, or call `get_new_bucket_name()` just to generate the name

cbodley: oh, the tenanted user would create a different bucket with the same name. i guess dbstore just doesn't handle the tenants correctly

tobias-urdin: ah nice catch, I've changed to your suggestion ``get_new_bucket(client)``

cbodley:
cc @alimaredia i cherry-picked this to ceph-pacific after merging https://github.com/ceph/ceph/pull/55550

it's still needed on ceph-master, but i'll wait until after the squid freeze in case there are pending jobs without the fix from https://github.com/ceph/ceph/pull/55458

cbodley:
cherry-picked to ceph-master

PR #547, opened 2024-02-09
s3: object lock tests for deletion of multipart objects

cbodley:
test cases for https://github.com/ceph/ceph/pull/54767 / https://tracker.ceph.com/issues/63724


cbodley:
ping @alimaredia

PR #552, opened 2024-02-20
replace datetime.now with datetime.utcnow()

alimaredia:
when the local timezone is not UTC and if it is a day behind, lifecycle_header tests fails with 2 days not equal to 1 so replacing datetime.now() with datetime.utcnow()

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -8588,7 +8588,7 @@ def test_lifecycle_expiration_header_put():
     bucket_name = get_new_bucket()
     client = get_client()
 
-    now = datetime.datetime.now(None)
+    now = datetime.datetime.utcnow()
     response = setup_lifecycle_expiration(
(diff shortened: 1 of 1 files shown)


alimaredia:
https://github.com/ceph/s3-tests/pull/550 fixes this issue on `ceph-pacific` but backports will still need to be made for `ceph-quincy`, and `ceph-reef`. There was an additional test on `master` that did not exist on `ceph-pacific` so I opened this PR instead of having that one go to `master` instead.

cbodley (approved):

alimaredia:
Run scheduled: https://pulpito.ceph.com/amaredia-2024-02-20_22:51:33-rgw:lifecycle-main-distro-default-smithi/

alimaredia:
Rerun: https://pulpito.ceph.com/amaredia-2024-02-21_05:36:14-rgw:lifecycle-main-distro-default-smithi/

alimaredia:
This commit has been backported to `ceph-master`, `ceph-quincy`, and `ceph-reef`.

PR #553, opened 2024-03-01
Add markers encryption_sse_c and encryption_sse_kms

atta:
* testing of sse-c is now possible without a configured kms
* add header 'X-Forwarded-Proto' = 'https' if is_secure is set to False in the configuration file

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -9,6 +9,8 @@ markers =
     bucket_encryption
     cloud_transition
     encryption
+    encryption_sse_c
+    encryption_sse_kms
     fails_on_aws
     fails_on_dbstore
     fails_on_dho
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -5171,7 +5171,7 @@ def test_list_buckets_invalid_auth():
     e = assert_raises(ClientError, bad_auth_client.list_buckets)
     status, error_code = _get_status_and_error_code(e.response)
     assert status == 403
-    assert error_code == 'InvalidAccessKeyId'
+    assert error_code == 'AccessDenied'
 
 def test_list_buckets_bad_auth():
     main_access_key = get_main_aws_access_key()
@@ -5179,7 +5179,7 @@ def test_list_buckets_bad_auth():
     e = assert_raises(ClientError, bad_auth_client.list_buckets)
     status, error_code = _get_status_and_error_code(e.response)
     assert status == 403
-    assert error_code == 'SignatureDoesNotMatch'
+    assert error_code == 'AccessDenied'
 
 @pytest.fixture
 def override_prefix_a():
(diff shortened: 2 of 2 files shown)


atta:
Hi, please let me know if there is anything to change, would love to get it done

LenzGr:
Hey @cbodley or @mattbenjamin - would you mind taking a look at this PR? Thank you :)

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,23 @@
...
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
+    - name: Analysing the code with pylint
+      run: |
+        pylint $(git ls-files '*.py')
```
cbodley, line 23: please remove this

pytest.ini
```diff
@@ -9,6 +9,8 @@ markers =
     bucket_encryption
     cloud_transition
     encryption
+    encryption_sse_c
+    encryption_sse_kms
```
cbodley, line 13: there's already a marker for `sse_s3`. please remove the `encryption_` part to be consistent

s3tests_boto3/functional/test_s3.py
```diff
@@ -5171,15 +5171,15 @@ def test_list_buckets_invalid_auth():
     e = assert_raises(ClientError, bad_auth_client.list_buckets)
     status, error_code = _get_status_and_error_code(e.response)
     assert status == 403
-    assert error_code == 'InvalidAccessKeyId'
+    assert error_code == 'AccessDenied'
```
cbodley, line 5174: what's the motivation of this change? afaik this has been passing as-is

s3tests_boto3/functional/test_s3.py
```diff
@@ -8607,6 +8607,8 @@ def _test_encryption_sse_customer_write(file_size):
         'x-amz-server-side-encryption-customer-key': 'pO3upElrwuEXSoFwCfnZPdSsmt/xWeFa0N9KgDijwVs=',
         'x-amz-server-side-encryption-customer-key-md5': 'DWygnHRtgiJ77HCm+1rvHw=='
     }
+    if get_config_is_secure() == False:
+        sse_client_headers['X-Forwarded-Proto'] = 'https'
```
cbodley, line 8611: this shouldn't be necessary, and servers probably shouldn't accept this header from clients without additional configuration. if you're just testing against ceph rgw, you can set rgw_crypt_require_ssl=false to run these

atta: In my tests, if the `rgw_crypt_require_ssl=false` is set, the ceph rgw will require the header in order to accept the request.

cbodley: that config option has always worked in our automated testing, ex. https://github.com/ceph/ceph/blob/04416f4/qa/suites/rgw/crypt/3-rgw/rgw.yaml#L7

PR #555, opened 2024-03-10
test_headers: use fixture to hook request headers

cbodley:
with pytest 8.1.1, the setup/teardown functions no longer appear to run. replace them with a pytest fixture

Fixes: https://tracker.ceph.com/issues/64822


cbodley:
cherry-picked to ceph-master, ceph-squid, ceph-reef

PR #561, opened 2024-04-17
Fix non handled error response

galsalomon66:
add handling for a new exception


cbodley (approved):

cbodley:
i've cherry-picked these to ceph-master

PR #564, opened 2024-05-01
Wip cbodley multipart nostreaming

mattbenjamin:


s3tests_boto3/functional/test_s3.py
```diff
@@ -13430,3 +13430,130 @@ def test_get_object_torrent():
...
+    body = FakeWriteFile(size, 'C')
+    part3_sha256sum = 'Vw7oB/nKQ5xWb3hNgbyfkvDiivl+U+/Dft48nfJfDow='
+    response = client.upload_part(UploadId=upload_id, Bucket=bucket, Key=key, PartNumber=3, Body=body, ChecksumAlgorithm='SHA256', ChecksumSHA256=part3_sha256sum)
+    etag3 = response['ETag'].strip('"')
+
+    composite_sha256sum = 'uWBwpe1dxI4Vw8Gf0X9ynOdw/SS6VBzfWm9giiv1sf4=-3'
+    response = client.complete_multipart_upload(Bucket=bucket, Key=key, UploadId=upload_id, ChecksumSHA256=composite_sha256sum, MultipartUpload={'Parts': [
+        {'ETag': etag1, 'ChecksumSHA256': response['ChecksumSHA256'], 'PartNumber': 1},
+        {'ETag': etag2, 'ChecksumSHA256': response['ChecksumSHA256'], 'PartNumber': 2},
+        {'ETag': etag3, 'ChecksumSHA256': response['ChecksumSHA256'], 'PartNumber': 3}]})
```
cbodley, line 13553: this is passing the same ChecksumSHA256 value for each part, which comes from the last upload_part() response. CompleteMultipart should check that these checksums match the parts and reject such a request

you should be able to pass part1_sha256sum etc for these values

s3tests_boto3/functional/test_s3.py
```diff
@@ -13430,3 +13430,130 @@ def test_get_object_torrent():
...
+    upload_id = response['UploadId']
+
+    size = 1024
+    body = FakeWriteFile(size, 'A')
+    part_sha256sum = 'arcu6553sHVAiX4MjW0j7I7vD4w6R+Gz9Ok0Q9lTa+0='
+    response = client.upload_part(UploadId=upload_id, Bucket=bucket, Key=key, PartNumber=1, Body=body, ChecksumAlgorithm='SHA256', ChecksumSHA256=part_sha256sum)
+
+    # should reject the missing part checksum
+    e = assert_raises(ClientError, client.complete_multipart_upload, Bucket=bucket, Key=key, UploadId=upload_id, ChecksumSHA256='bad', MultipartUpload={'Parts': [
+        {'ETag': response['ETag'].strip('"'), 'PartNumber': 1}]})
```
cbodley, line 13498: oops, i'd made a mistake here - above we already tested that `ChecksumSHA256='bad'` leads to `400 InvalidRequest`

this one was meant to test the case where ChecksumSHA256 is missing from the Parts array

`ChecksumSHA256='bad'` here should pass the correct value `ChecksumSHA256=composite_sha256sum` (as below) so that we correctly test the Parts behavior

since test_multipart_checksum_3parts is passing with the wrong checksum values in Parts, i think https://github.com/ceph/ceph/pull/55076 is missing the logic to validate those

mattbenjamin: oh, true;  it validates the checksums of the parts as actually uploaded, though

s3tests_boto3/functional/test_s3.py
```diff
@@ -13430,3 +13430,182 @@ def test_get_object_torrent():
...
         assert error_code == 'NoSuchKey'
+
+@pytest.mark.checksum
+def test_object_checksum_sha256():
+    bucket = get_new_bucket()
+    client = get_client()
+
+    key = "myobj"
+    size = 1024
+    body = FakeWriteFile(size, 'A')
```
rkhudov, line 13441:
```suggestion
    body = FakeWriteFile(1024, 'A')
```
Since it is not used inside the test except this place, can some some memory to not assign a variable to it :)

s3tests_boto3/functional/test_s3.py
```diff
@@ -13430,3 +13430,182 @@ def test_get_object_torrent():
...
+    # 2024-04-25T17:47:47.991-0400 7f78e3a006c0  2 req 4931907640780566174 0.011000143s s3:complete_multipart completing
+    # 2024-04-25T17:47:47.991-0400 7f78e3a006c0  1 req 4931907640780566174 0.011000143s s3:complete_multipart ERROR: either op_ret is negative (execute failed) or target_obj is null, op_ret: -2200
+    # -2200 turns into 500, InternalError
+
+    key = "mymultipart2"
+    response = client.create_multipart_upload(Bucket=bucket, Key=key, ChecksumAlgorithm='SHA256')
+    assert 'SHA256' == response['ChecksumAlgorithm']
+    upload_id = response['UploadId']
+
+    size = 1024
```
rkhudov, line 13490:
```suggestion
```
Can use `size` variable from [13466](https://github.com/ceph/s3-tests/pull/564/files#diff-7c5f59564fac3971102a77540c44877c2d255783d45893452b24b696632ca088R13466) line

s3tests_boto3/functional/test_s3.py
```diff
@@ -13430,3 +13430,182 @@ def test_get_object_torrent():
...
+    status, error_code = _get_status_and_error_code(e.response)
+    assert status == 400
+    assert error_code == 'InvalidRequest'
+
+    key = "mymultipart3"
+    response = client.create_multipart_upload(Bucket=bucket, Key=key, ChecksumAlgorithm='SHA256')
+    assert 'SHA256' == response['ChecksumAlgorithm']
+    upload_id = response['UploadId']
+
+    size = 1024
```
rkhudov, line 13507:
```suggestion
```
Same here

rkhudov (commented):
Sorry, I might be not the right person to review, but just wanted to add small comments

cbodley:
> mattbenjamin wants to merge 200 commits

i think something went wrong with the rebase here

cbodley (approved):

cbodley:
i've cherry-picked these to ceph-master

cbodley:
cherry-picked to ceph-squid for https://github.com/ceph/ceph/pull/62312

PR #567, opened 2024-05-22
BucketPolicy: add tests for ConfirmRemoveSelfBucketAccess header

clwluvw:
Refrence: https://github.com/ceph/ceph/pull/57629

### DIFF
--- a/s3tests_boto3/functional/__init__.py
+++ b/s3tests_boto3/functional/__init__.py
@@ -804,3 +804,24 @@ def get_restore_debug_interval():
 
 def get_read_through_days():
     return config.read_through_restore_days
+
+def create_iam_user_s3client(client):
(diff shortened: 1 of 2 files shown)


s3tests_boto3/functional/__init__.py
```diff
@@ -804,3 +804,24 @@ def get_restore_debug_interval():
 
 def get_read_through_days():
     return config.read_through_restore_days
+
+def get_iam_user_s3client(client):
```
cbodley, line 808: most of these `get_foo_client()` functions are just constructing a boto client with existing creds. please rename get_ to create_ so callers know there's something to clean up

clwluvw: done.

cbodley (approved):

PR #570, opened 2024-06-26
test Get/HeadObject with partNumber for single-multipart upload

cbodley:
test_multipart_get_part() tests 'normal' multipart uploads. add a new test case for a multipart upload with a single part to tests the fix for https://tracker.ceph.com/issues/66705 in https://github.com/ceph/ceph/pull/58288


cbodley:
cherry-picked to ceph-squid

PR #573, opened 2024-07-19
BucketPolicy: donot allow NotPrincipal with Allow Effect

clwluvw:
Ref. https://github.com/ceph/ceph/pull/58686


cbodley (approved):

cbodley:
cherry-picked to ceph-master

PR #574, opened 2024-07-19
BucketPolicy: add test for sse-c in conditions

clwluvw:
Ref. https://github.com/ceph/ceph/pull/58689

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -10084,12 +10084,6 @@ def test_encryption_sse_c_enforced_with_bucket_policy():
     bucket_name = get_new_bucket()
     client = get_client()
 
-    deny_incorrect_algo = {
-        "StringNotEquals": {
-          "s3:x-amz-server-side-encryption-customer-algorithm": "AES256"
-        }
-    }
-
     deny_unencrypted_obj = {
         "Null" : {
           "s3:x-amz-server-side-encryption-customer-algorithm": "true"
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/test_s3.py
```diff
@@ -10077,6 +10077,44 @@ def test_encryption_sse_c_post_object_authenticated_request():
...
+
+    deny_incorrect_algo = {
+        "StringNotEquals": {
+          "s3:x-amz-server-side-encryption-customer-algorithm": "AES256"
+        }
+    }
+
+    deny_unencrypted_obj = {
+        "Null" : {
+          "s3:x-amz-server-side-encryption-customer-algorithm": "true"
```
cbodley, line 10095: is this first condition relevant to the test? it looks like the `check_access_denied` part only depends on the second

clwluvw: That's my confusion part as well... Apparantly the stringNotEqual doesn't deny if the header is not specified at all (IIRC same for sse-s3) and the string match happens when it's there (but the only option is AES256). Maybe I can decouple it via status code 403 vs 400.

cbodley: i guess that's the difference between StringNotEquals and StringNotEqualsIfExists? from https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html#Conditions_IfExists
> If you are using an "Effect": "Deny" element with a negated condition operator like StringNotEqualsIfExists, the request is still denied even if the tag is missing.

for the purposes of testing https://github.com/ceph/ceph/pull/58689, i tend to think the Null condition is sufficient to verify that we understand that condition key. what do you think?

clwluvw: @cbodley Can you please check my last commit? I guess this can cover the case to make sure the value is also respected by RGW. But if you think it's unnecessary I'll drop it. np.

cbodley: update looks great

PR #576, opened 2024-08-08
rgw/logging: basic set of tests for bucket logging

yuvalif:
bucket logging is implemented here:  https://github.com/ceph/ceph/pull/59808
~~https://github.com/ceph/ceph/pull/54815~~

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -100,3 +100,17 @@ You can filter tests based on their attributes::
 
 	S3TEST_CONF=your.conf tox -- s3tests_boto3/functional/test_iam.py -m 'not fails_on_rgw'
 
+========================
+ Bucket logging tests
+========================
+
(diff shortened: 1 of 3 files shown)


yuvalif:
* add versioned bucket tests

yuvalif:
test is passing. see: https://github.com/ceph/ceph/pull/60700#issuecomment-2493444419

cbodley:
why is it necessary to duplicate the s3 extensions file from the ceph repo? i'd rather not worry about trying to keep those in sync

yuvalif:
> why is it necessary to duplicate the s3 extensions file from the ceph repo? i'd rather not worry about trying to keep those in sync

the other (and better) option is to fetch it from the ceph repo, but in my case I had to have a way to fetch it from a branch on my ceph fork.

cbodley:
cherry-picked to ceph-master

PR #577, opened 2024-08-13
s3: reenable tenanted bucket policy test

cbodley:
the before-call hook url-encodes the ':' part of tenanted bucket names to resolve SignatureDoesNotMatch errors

removed the list-v2 version of the test since it isn't relevant to bucket policy test coverage

add a new test case that creates the bucket under the tenanted user, then uses the main client to access it

### DIFF
--- a/s3tests_boto3/functional/__init__.py
+++ b/s3tests_boto3/functional/__init__.py
@@ -259,6 +259,7 @@ def configure():
     config.tenant_display_name = cfg.get('s3 tenant',"display_name")
     config.tenant_user_id = cfg.get('s3 tenant',"user_id")
     config.tenant_email = cfg.get('s3 tenant',"email")
+    config.tenant_name = cfg.get('s3 tenant',"tenant")
 
     config.iam_access_key = cfg.get('iam',"access_key")
     config.iam_secret_key = cfg.get('iam',"secret_key")
@@ -693,6 +694,9 @@ def get_tenant_aws_secret_key():
 
 def get_tenant_display_name():
     return config.tenant_display_name
+
+def get_tenant_name():
+    return config.tenant_name
 
 def get_tenant_user_id():
     return config.tenant_user_id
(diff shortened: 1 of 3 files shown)


cbodley:
the existing test case `test_bucket_policy_different_tenant()` passes against rgw

> add a new test case that creates the bucket under the tenanted user, then uses the main client to access it

this new test case `test_bucket_policy_tenanted_bucket()` fails against rgw without https://github.com/ceph/ceph/pull/59169, and passes with 👍

pritha-srivastava (approved):

s3tests_boto3/functional/__init__.py
```diff
@@ -694,6 +694,10 @@ def get_tenant_aws_secret_key():
 def get_tenant_display_name():
     return config.tenant_display_name
 
+def get_tenant_name():
+    s = config.tenant_display_name
+    return s[:s.index('$')]
```
cbodley, line 699: this works for the default value from s3tests.conf.SAMPLE:
```ini
[s3 tenant]
# tenant display_name set in vstart.sh
display_name = testx$tenanteduser
```
but fails in teuthology where [qa/tasks/s3tests.py](https://github.com/ceph/ceph/blob/53b533f/qa/tasks/s3tests.py#L82) sets display_name to `Mr.testx$tenanteduser`. rgw sees these requests as:
```
REQUEST_URI=/Mr.testx%3Atest-fak34zuywqcog5fhyh4e22bc-475?encoding-type=url
```
and rejects it with `400 InvalidTenantName` because of the `.` in `Mr.testx`

cbodley: since we need the tenant name for this test new case, i added a new `tenant` variable to the `[s3 tenant]` config section. i added a new commit to https://github.com/ceph/ceph/pull/59169 that configures this in teuthology

cbodley:
cherry-picked to ceph-master

PR #579, opened 2024-08-19
s3select: fix the assert per empty results

galsalomon66:
s3select tests use an assert on 2 statement results
the 2 statements have different syntax but should return the same results for the same input (its semantically equal).

upon an empty result on both statements, the test infrastructure failed on assert-false.

the PR changes this behavior, in case both are empty, its OK, with a warning message.
https://tracker.ceph.com/issues/65651

### DIFF
--- a/s3tests_boto3/functional/test_s3select.py
+++ b/s3tests_boto3/functional/test_s3select.py
@@ -121,7 +121,7 @@ def s3select_assert_result(a,b):
             stack = traceback.extract_stack(limit=2)
             formatted_stack = traceback.format_list(stack)[0]
             warnings.warn(UserWarning("{}".format(formatted_stack)))
-            return a==b
+            return True
         assert a_strip != ""
         assert b_strip != ""
     else:
@@ -131,10 +131,10 @@ def s3select_assert_result(a,b):
             stack = traceback.extract_stack(limit=2)
             formatted_stack = traceback.format_list(stack)[0]
             warnings.warn(UserWarning("{}".format(formatted_stack)))
-            return a==b
+            return True
         assert a != ""
         assert b != ""
-    assert a == b
+    assert True
 
 def create_csv_object_for_datetime(rows,columns):
         result = ""
--- a/s3tests_boto3/functional/test_s3select.py
+++ b/s3tests_boto3/functional/test_s3select.py
@@ -7,6 +7,8 @@ from botocore.exceptions import ClientError
 from botocore.exceptions import EventStreamError
 
 import uuid
+import warnings
+import traceback
 
 from . import (
     configfile,
@@ -96,6 +98,7 @@ def test_generate_where_clause():
     for _ in range(100): 
         generate_s3select_where_clause(bucket_name,obj_name)
 
+
 @pytest.mark.s3select
 def test_generate_projection():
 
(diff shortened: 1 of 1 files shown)


galsalomon66:
https://pulpito.ceph.com/gsalomon-2024-08-20_13:17:09-rgw:verify-s3test_handle_empty_results-distro-default-smithi/
24 jobs
all green

s3tests_boto3/functional/test_s3select.py
```diff
@@ -112,9 +115,23 @@ def s3select_assert_result(a,b):
     if type(a) == str:
         a_strip = a.strip()
         b_strip = b.strip()
+        if a=="" and b=="":
+            warnings.warn(UserWarning("{}".format("both results are empty, it may indicates a wrong input, please check the test input")))
+            ## print the calling function that created the empty result.
+            stack = traceback.extract_stack(limit=2)
+            formatted_stack = traceback.format_list(stack)[0]
+            warnings.warn(UserWarning("{}".format(formatted_stack)))
+            return a==b
```
ivancich, line 124: Could this be replaced with `return True`? It seems that it's been established that `a` and `b` are the same, unless I'm missing something subtle in Python.

s3tests_boto3/functional/test_s3select.py
```diff
@@ -112,9 +115,23 @@ def s3select_assert_result(a,b):
...
         assert a_strip != ""
         assert b_strip != ""
     else:
+        if a=="" and b=="":
+            warnings.warn(UserWarning("{}".format("both results are empty, it may indicates a wrong input, please check the test input")))
+            ## print the calling function that created the empty result.
+            stack = traceback.extract_stack(limit=2)
+            formatted_stack = traceback.format_list(stack)[0]
+            warnings.warn(UserWarning("{}".format(formatted_stack)))
+            return a==b
```
ivancich, line 134: same as above

ivancich (approved):

s3tests_boto3/functional/test_s3select.py
galsalomon66: nothing important.
just explaining this flow, 
assert is TRUE since result-a is equal to result-b.

cbodley:
@galsalomon66 i tried to cherry-pick this to the ceph-master branch, but hit the following conflict:
```diff
@@@ -794,7 -821,11 +811,15 @@@ def test_true_false_in_expressions()
      csv_obj = create_random_csv_object(10000,10)

      csv_obj_name = get_random_string()
++<<<<<<< HEAD
 +    bucket_name = "test"
++=======
+     bucket_name = get_new_bucket_name()
+
+     ## 1,2 must exist in first/second column (to avoid empty results)
+     csv_obj = csv_obj + "1,2,,,,,,,,,,\n"
+
++>>>>>>> 9444c29 (fix the assert per empty results)
      upload_object(bucket_name,csv_obj_name,csv_obj)

      res_s3select_in = remove_xml_tags_from_result(  run_s3select(bucket_name,csv_obj_name,'select int(_1) from s3object where (int(_1) in(1)) = true;')).rep
lace("\n","")
```
it looks like ceph-master is missing changes from https://github.com/ceph/s3-tests/pull/513. but i see a later unmerged pr https://github.com/ceph/s3-tests/pull/518 to revert that stuff. any idea what's going on there?

galsalomon66:
> thanks @galsalomon66. the commits from #513 don't cherry-pick cleanly either. could i ask you to prepare a pr against ceph-master that backports all of the necessary changes?
> 
> (ceph's teuthology tests run against the ceph-master branch so we need to keep it up-to-date)

https://github.com/ceph/s3-tests/pull/587

PR #589, opened 2024-09-25
Add v2 signature presigned get_object tests

tobias-urdin:
This adds tests for get_object presigned URLs
using signature v2.

Also code formatting.

This change depends on [1].

[1] https://github.com/ceph/ceph/pull/59977

PR #590, opened 2024-09-25
s3: add sigv2 tests for cors presigned urls

cbodley:


cbodley:
closing in favor of https://github.com/ceph/s3-tests/pull/589

PR #591, opened 2024-09-27
s3: test GetObject with PartNumber and SSE-C encryption

cbodley:
test case for https://github.com/ceph/ceph/pull/60019

copies the existing `test_multipart_get_part()` test but replaces `_multipart_upload()` with `_multipart_upload_enc()`


s3tests_boto3/functional/test_s3.py
```diff
@@ -6459,6 +6459,61 @@ def test_multipart_get_part():
...
+        assert response['PartsCount'] == part_count
+        assert response['ETag'] == '"{}"'.format(part['ETag'])
+
+        response = client.get_object(Bucket=bucket_name, Key=key, PartNumber=part['PartNumber'], **get_args)
+        assert response['PartsCount'] == part_count
+        assert response['ETag'] == '"{}"'.format(part['ETag'])
+        assert response['ContentLength'] == size
+        # compare contents
+        for chunk in response['Body'].iter_chunks():
+            assert chunk.decode() == data[0:len(chunk)]
```
cbodley, line 6508: when this test runs against rgw without the fix from https://github.com/ceph/ceph/pull/60019, it fails here with:
> UnicodeDecodeError: 'utf-8' codec can't decode byte 0x84 in position 4: invalid start byte

because it's getting the encrypted bytes

cbodley:
cherry-picked to ceph-master

PR #594, opened 2024-10-09
s3tests_boto3/functional/test_s3.py: add test for list object versions with key marker

samakshd:
The key-marker parameter in the list-object-versions operation is inclusive in RGW, while it's exclusive in S3. This commit adds the test to check the corresponding case.

Issue ref: https://tracker.ceph.com/issues/68055
Fix PR: https://github.com/ceph/ceph/pull/60068

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -1450,6 +1450,22 @@ def test_bucket_list_return_data_versioning():
         assert obj['VersionId'] == key_data['VersionId']
         _compare_dates(obj['LastModified'],key_data['LastModified'])
 
+@pytest.mark.fails_on_dbstore
+def test_bucket_list_return_data_versioning_key_marker():
+
+    bucket_name = get_new_bucket()
+    check_configure_versioning_retry(bucket_name, "Enabled", "Enabled")
+    key_names = ['bar', 'baz', 'foo']
+    bucket_name = _create_objects(bucket_name=bucket_name,keys=key_names)
+
+    client = get_client()
+    response  = client.list_object_versions(Bucket=bucket_name, KeyMarker='baz')
+    assert response['KeyMarker'] == 'baz'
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/test_s3.py
```diff
@@ -1450,6 +1450,22 @@ def test_bucket_list_return_data_versioning():
         assert obj['VersionId'] == key_data['VersionId']
         _compare_dates(obj['LastModified'],key_data['LastModified'])
 
+@pytest.mark.fails_on_dbstore
```
cbodley, line 1453: i don't think this is needed - there are other `list_object_versions()` tests that work with dbstore

cbodley (commented):
thanks, this looks good. but i also don't see any tests for `list_object_versions()` with `VersionIdMarker`. could you please write two versions of the same object name and verify that `VersionIdMarker` is also inclusive?

samakshd:
> thanks, this looks good. but i also don't see any tests for `list_object_versions()` with `VersionIdMarker`. could you please write two versions of the same object name and verify that `VersionIdMarker` is also inclusive?

@cbodley, I wanted to know can there also be scenario where just versionIdMarker is passed and keyMarker is not passed?

cbodley:
@samakshd not sure, but it probably doesn't do anything without KeyMarker

samakshd:
Hi @cbodley, I have added the tests for list object version with key marker and version id marker. I have also modified key marker test to make it more stronger as I realized that my initial PR fix was not correct. I have added another commit to my existing PR. 

However while rebasing my branch with updated main branch, I introduced some commits from other people in my PR only, this created kind of mess. Since I am new to use git this much extensively, this happened. I tried resetting my branch to first commit, by git hard reset, and then did fresh changes to it and merged first with updated main branch and then pushed it. This however did not revert some review requests and tags that got into PR due to the commits that came due to rebasing.

cbodley (approved):

PR #598, opened 2024-10-31
DeleteObjects: ensure 400 MalformedXML is returned for empty list

clwluvw:
Ref: https://github.com/ceph/ceph/pull/60579


clwluvw:
@ivancich Thanks for QAing the original PR. I'm afraid I forgot to mention this test case boldly in the main PR, and you might have missed this running. Can you please have a look and maybe run it by any next QA to make sure it's not breaking anything, and we let this go through as well?

PR #602, opened 2024-11-20
rgw/s3tests: making the tests generic to check only

pritha-srivastava:
for 403 error code (and not specific errors) so that the tests pass for any order of auth engines.


cbodley:
cherry-picked to ceph-master

cbodley:
cherry-picked to ceph-squid

PR #604, opened 2024-11-21
sts: remove test_get_session_token_permanent_creds_denied

cbodley:
this test fails if `rgw_s3_auth_order` puts local before sts, because the local engine will gladly accept the s3_main user's credentials

Fixes: https://tracker.ceph.com/issues/69001

### DIFF
--- a/s3tests_boto3/functional/test_sts.py
+++ b/s3tests_boto3/functional/test_sts.py
@@ -183,39 +183,6 @@ def test_get_session_token():
 
 @pytest.mark.test_of_sts
 @pytest.mark.fails_on_dbstore
-def test_get_session_token_permanent_creds_denied():
-    s3bucket_error=None
-    iam_client=get_iam_client()
-    sts_client=get_sts_client()
-    sts_user_id=get_alt_user_id()
-    default_endpoint=get_config_endpoint()
-    s3_main_access_key=get_main_aws_access_key()
-    s3_main_secret_key=get_main_aws_secret_key()
-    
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/test_sts.py
```diff
@@ -200,8 +198,6 @@ def test_get_session_token_permanent_creds_denied():
     assert response['ResponseMetadata']['HTTPStatusCode'] == 200
     
     s3_client=boto3.client('s3',
-                aws_access_key_id = s3_main_access_key,
-		aws_secret_access_key = s3_main_secret_key,
```
cbodley, line 204: hmm, this causes the `create_bucket()` call to fail with
> botocore.exceptions.NoCredentialsError: Unable to locate credentials

pritha-srivastava: @cbodley : As part of STSLite (predecessor to STS) in rgw, `getsessiontoken` was implemented to reduce the load on Keystone/LDAP or any external server. The implementation was AWS non-compliant and the idea was to authenticate `GetSessionToken` with the external server using the EC2/LDAP creds and return temporary creds, and these temp creds could be used for further s3 ops.  This idea was extended to Local Auth also, hence this test case covers a failure scenario for local auth. I do not think anybody is using/used STS Lite, so we can either ignore this failure, mark this as fails_on_rgw or just remove it.

pritha-srivastava (approved):

cbodley:
cherry-picked to ceph-master

PR #607, opened 2025-01-02
rgw/logging: increase test coverage

yuvalif:


yuvalif:
bucket logging test suite is passing: https://pulpito.ceph.com/yuvalif-2025-01-07_05:43:12-rgw:bucket-logging-wip-yuval-logging-tenant-distro-default-smithi/

yuvalif:
@AliMasarweh , @cbodley can you please help review this (ceph PR already merged)?

s3tests_boto3/functional/test_s3.py
```diff
@@ -14319,6 +14319,14 @@ def test_put_bucket_logging():
     assert response['LoggingEnabled'] == logging_enabled
 
 
+def _flush_logs(client, src_bucket_name, dummy_key="dummy"):
+    if _has_bucket_logging_extension():
+        client.post_bucket_logging(Bucket=src_bucket_name)
+    else:
+        time.sleep(expected_object_roll_time*1.1)
+        client.put_object(Bucket=src_bucket_name, Key=dummy_key, Body='dummy')
```
AliMasarweh, line 14327: If it has bucket logging enabled we do a POST operation to flush it,
Else if bucket logging disabled, we put an object called "dummy", can you explain this? is this the old way of flushing before it was supported in REST api?

yuvalif: yes. this is for clients that dont have our extensions.
the way to "flush", is to wait for the timeout, and then send another put request to trigger the rollover

AliMasarweh (approved):
LGTM

PR #608, opened 2025-01-10
Switch from ssl.wrap_socket to SSLContext.wrap_socket

jheunis-bloomberg:
The python docs specify that since python 3.2, it is recommended to use `SSLContext.wrap_socket` instead of `ssl.wrap_socket` ("since it is both inefficient and has no support for SNI and hostname matching").

In python3.12, `ssl.wrap_socket` was removed, which causes the 100-continue test to reliably fail whenever it's configured with `is_secure=True`. Switching to the recommended `SSLContext.wrap_socket` (using the default SSL context) fixes this, allowing the test to run cleanly again.

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -6799,7 +6799,8 @@ def _simple_http_req_100_cont(host, port, is_secure, method, resource):
 
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     if is_secure:
-        s = ssl.wrap_socket(s);
+        ssl_context = ssl.create_default_context()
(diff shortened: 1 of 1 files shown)


adamemerson (approved):

jmundack:
@adamemerson  - thanks!
anything else needed here before merge?

adamemerson:
Schedule a test run against this branch on Teuthology, you'll need to make a custom branch of Ceph that points to this branch of s3-tests.

jheunis-bloomberg:
Independently fixed by Casey in https://github.com/ceph/s3-tests/pull/734 and ported along with the rest of the code to the ceph repo.

PR #615, opened 2025-02-03
squid is missing exception handling, it cause s3-tests to fail

galsalomon66:
the squid branch is missing the stream-error exception handling.
it causes s3-tests to failed.


adamemerson:
I apologize, but I ended up having to revert this as it was messing up Squid QA. Could you do a test run against it and see what's up?

PR #625, opened 2025-02-26
s3: test bucket policy evaluation for CreateMultipartUpload

cbodley:
test case for https://tracker.ceph.com/issues/70191


cbodley:
cherry-picked to ceph-master

PR #640, opened 2025-04-08
Allow exceptions to pass when cleaning up buckets

dang:
We clean up buckets 3 times, using each client, so that all buckets can be removed, regardless of which client created it.  However, the cleanup will throw an exception if permission is denied, due to using the wrong
user.   This will stop cleanup, and cause all subsequent tests to ERROR.
Instead, pass these exceptions, allowing full cleanup.

### DIFF
--- a/s3tests_boto3/functional/__init__.py
+++ b/s3tests_boto3/functional/__init__.py
@@ -305,9 +305,21 @@ def setup():
 def teardown():
     alt_client = get_alt_client()
     tenant_client = get_tenant_client()
-    nuke_prefixed_buckets(prefix=prefix)
-    nuke_prefixed_buckets(prefix=prefix, client=alt_client)
-    nuke_prefixed_buckets(prefix=prefix, client=tenant_client)
+    try:
+        print('teardown: nuking with default client')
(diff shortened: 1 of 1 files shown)


cbodley:
i guess i'm conflicted on this. usually, i only see cleanup permission errors where something in the test case failed, so didn't run the logic to restore permissions. i view this as a bug in the test case itself. https://github.com/ceph/s3-tests/pull/428 is an example where we added `try/finally` for unconditional cleanup

if the cleanup logic ignores all errors, we risk introducing new test cases that silently leave things behind. i think i'd prefer to leave those as failures

dang:
The test case I'm trying to fix is test_object_raw_get_x_amz_expires_not_expired_tenant.  It creates a bucket with "public-read" and an object with "public-read" using the tenant_client.  Then, teardown tries (first) do delete with the normal client, and fails, and aborts.  I don't see how this test can ever pass.

dang:
The alternative would be to not ERROR test cases unless *that test case* left things behind.  The issue is that all the rest of the test cases ERROR for the rest of the run.

PR #641, opened 2025-04-17
rgw/logging: mark test that require rollover time set

yuvalif:
this is adding needed marker for this ceph PR: https://github.com/ceph/ceph/pull/61486


cbodley:
cherry-picked to ceph-master

PR #647, opened 2025-05-07
rgw/logging: add tests for flushed object name

yuvalif:
when using API call to flush the pending logging object of the bucket, the flshed object name is returned.
tests were added to verify the right name is returned.


yuvalif:
test run: https://github.com/ceph/ceph/pull/63160#issuecomment-2939974674

PR #650, opened 2025-05-12
rgw/logging: test mpu complete in standard mode

yuvalif:
this test is used to verify: 
https://github.com/ceph/ceph/pull/63244
https://github.com/ceph/ceph/pull/63268


yuvalif:
test is passing. see: https://github.com/ceph/ceph/pull/63268#issuecomment-2911057627

PR #651, opened 2025-05-13
rgw/logging: test log object name order

yuvalif:
verify that the names of the log objects in the log bucket are lexicographically ordered.

this is test for: https://github.com/ceph/ceph/pull/63264


yuvalif:
test is passing. see: https://github.com/ceph/ceph/pull/63264#issuecomment-2911075704

PR #659, opened 2025-06-04
pytest: use parametrize for s3website test_routing_generator()

cbodley:
resolves error with recent pytest:
```
==================================== ERRORS ====================================
____________ ERROR collecting s3tests/functional/test_s3_website.py ____________
'yield' keyword is allowed in fixtures, but not in tests (test_routing_generator)
```
use https://docs.pytest.org/en/7.1.x/example/parametrize.html instead to generate a `test_routing_generator()` for each entry in `ROUTING_RULES_TESTS`

### DIFF
--- a/s3tests/functional/test_s3_website.py
+++ b/s3tests/functional/test_s3_website.py
@@ -1077,8 +1077,8 @@ def routing_check(*args, **kwargs):
 @pytest.mark.s3website_routing_rules
 @pytest.mark.s3website
 @pytest.mark.fails_on_dbstore
-def test_routing_generator():
-    for t in ROUTING_RULES_TESTS:
-        if 'xml' in t and 'RoutingRules' in t['xml'] and len(t['xml']['RoutingRules']) > 0:
-            t['xml']['RoutingRules'] = common.trim_xml(t['xml']['RoutingRules'])
-        yield routing_check, t
+@pytest.mark.parametrize('t', ROUTING_RULES_TESTS)
+def test_routing_generator(t):
+    if 'xml' in t and 'RoutingRules' in t['xml'] and len(t['xml']['RoutingRules']) > 0:
+        t['xml']['RoutingRules'] = common.trim_xml(t['xml']['RoutingRules'])
+    routing_check(t)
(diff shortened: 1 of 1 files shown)


yuvalif (approved):

cbodley:
https://qa-proxy.ceph.com/teuthology/cbodley-2025-06-04_20:17:31-rgw:website-main-distro-default-smithi/8310116/teuthology.log showed failures, which i'm hoping are resolved by the second commit

s3tests/functional/test_s3_website.py
```diff
@@ -1002,8 +1002,8 @@ def test_website_xredirect_private_abs():
 # give an error of 'The provided HTTP redirect code (314) is not valid. Valid codes are 3XX except 300.' during setting the website config
 # we should check that we can return that too on ceph
 
+@pytest.fixture
 def routing_setup():
-  check_can_test_website()
```
robbat2, line 1006: This is no longer called anywhere; and means there are cases where it can't be tested that are wrongly tested

cbodley: https://github.com/ceph/s3-tests/pull/482/commits/c80e9d2118ca5bcdec5285898a532f8e18faa6b6 made `check_can_test_website()` an auto-use fixture which applies to all test cases in the file:
```python
@pytest.fixture(autouse=True, scope="module")
def check_can_test_website():
```
`routing_setup()` doesn't need to call it now because it's already being called. i can confirm that it's still taking effect, because these are all skipped when run against a default vstart cluster:
```
s3tests/functional/test_s3_website.py::test_routing_generator[t0] SKIPPED (rgw_enable_static_website is false)
s3tests/functional/test_s3_website.py::test_routing_generator[t1] SKIPPED (rgw_enable_static_website is false)
s3tests/functional/test_s3_website.py::test_routing_generator[t2] SKIPPED (rgw_enable_static_website is false)
```

cbodley:
https://pulpito.ceph.com/cbodley-2025-06-05_04:04:11-rgw-main-distro-default-smithi/ is a mess but no s3-tests failures

cbodley:
cherry-picked to ceph-reef/squid/tentacle/master to resolve https://tracker.ceph.com/issues/71567

PR #663, opened 2025-06-27
s3: test conditional put_object() and delete_object()

cbodley:
test coverage for https://github.com/ceph/ceph/pull/63348
- [x] put_object() with IfMatch and IfNoneMatch
- [x] delete_object() with IfMatch
- [x] delete_object() with IfMatchLastModifiedTime and IfMatchSize
- [x] delete_objects() with `<ETag>`
- [x] delete_objects() with `<LastModifiedTime>`
- [x] delete_objects() with `<Size>`

put_object tests are duplicated to test non-versioned and overwrites of a current version
delete_object(s) tests are duplicated to test non-versioned, deletes of a current version, and deletes of a specific version

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -18314,6 +18314,8 @@ def test_put_object_if_match():
     assert (412, 'PreconditionFailed') == _get_status_and_error_code(e.response)
     e = assert_raises(ClientError, client.put_object, Bucket=bucket, Key=key, IfNoneMatch=etag)
     assert (412, 'PreconditionFailed') == _get_status_and_error_code(e.response)
+    response = client.put_object(Bucket=bucket, Key=key, IfNoneMatch='badetag')
+    assert 200 == response['ResponseMetadata']['HTTPStatusCode']
 
     client.put_object(Bucket=bucket, Key=key, IfMatch=etag)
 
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/test_s3.py
```diff
@@ -18001,7 +18001,7 @@ def test_put_object_if_match():
     etag = client.put_object(Bucket=bucket, Key=key, IfNoneMatch='*')['ETag']
 
     e = assert_raises(ClientError, client.put_object, Bucket=bucket, Key=key, IfNoneMatch='*')
-    assert (412, 'PreconditionFailed') == _get_status_and_error_code(e.response)
+    assert (404, 'NoSuchKey') == _get_status_and_error_code(e.response)
```
cbodley, line 18004: @AliMasarweh can you help me understand this change? i don't see any mention of 404 under [Conditional writes with If-None-Match header](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html#conditional-error-response), only for If-Match

AliMasarweh: if we have some object, but we pass **if-none-match="*"**,  
we return **-ENOENT** instead of **-ERR_PRECONDITION_FAILED**
I remember you told me that this what want to return

cbodley: one of us was probably confusing that with If-Match?

cbodley: the test checks for 412 because of this text:
> If there's an existing object, the write operation fails, resulting in a 412 Precondition Failed response.

AliMasarweh: done!

s3tests_boto3/functional/test_s3.py
```diff
@@ -3172,8 +3172,8 @@ def test_put_object_ifmatch_nonexisted_failed():
     client.meta.events.register('before-call.s3.PutObject', lf)
     e = assert_raises(ClientError, client.put_object, Bucket=bucket_name, Key='foo', Body='bar')
     status, error_code = _get_status_and_error_code(e.response)
-    assert status == 412
-    assert error_code == 'PreconditionFailed'
+    assert status == 404
+    assert error_code == 'NoSuchKey'
```
AliMasarweh, line 3176: fixed this failure from teuthology run

s3tests_boto3/functional/test_s3.py
```diff
@@ -18167,10 +18165,11 @@ def test_delete_object_current_if_match_last_modified_time():
     assert (412, 'PreconditionFailed') == _get_status_and_error_code(e.response)
 
     response = client.delete_object(Bucket=bucket, Key=key, IfMatchLastModifiedTime=mtime)
-    assert response['DeleteMarker']
+    assert 'DeleteMarker' in response
```
cbodley, line 18168: why change these? it would succeed even if the server returns `x-amz-delete-marker: false`

cbodley (commented):
ack on delete changes from 404->204 :+1:

s3tests_boto3/functional/test_s3.py
AliMasarweh: you are right, not sure what was I thinking

cbodley:
pushed to ceph-master as well

cbodley:
cherry-picked to ceph-tentacle

PR #667, opened 2025-07-01
bucket-logging: test for log record fields

nbalacha:
Tests that the AclRequired field is set correctly in the bucket logging record.

PR #670, opened 2025-07-04
bucket-logging: test the AuthType log field

nbalacha:
Adds a test to check the value of the AuthType field.

### DIFF
--- a/s3tests_boto3/functional/test_s3.py
+++ b/s3tests_boto3/functional/test_s3.py
@@ -14942,6 +14942,19 @@ def _verify_records(records, bucket_name, event_type, src_keys, record_type, exp
     return len(keys_found) == expected_count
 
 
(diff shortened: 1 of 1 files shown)


s3tests_boto3/functional/test_s3.py
```diff
@@ -14942,12 +14942,27 @@ def _verify_records(records, bucket_name, event_type, src_keys, record_type, exp
     return len(keys_found) == expected_count
 
 
+def _verify_record_field(records, bucket_name, event_type, object_key, record_type, field_name, expected_value):
```
nbalacha, line 14945: This has also been defined in https://github.com/ceph/s3-tests/pull/667. This PR will need to be modified once that one is merged.

nbalacha:
> Pending: Add a test for presigned URLs.
Done.

PR #672, opened 2025-07-14
delete, rgwlc: delete marker tests

mattbenjamin:
0. non-creation in non-versioned buckets
1. creation in versioned buckets
2. creation in versioning-suspended buckets
3. delete-marker expiration (lifecycle)

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -54,3 +54,4 @@ markers =
     user_policy
     versioning
     webidentity_test
+    delete_marker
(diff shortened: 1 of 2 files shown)


mattbenjamin:
(n.b., I'm going to refactor these a bit, not checking for both conditions [404 and header], just header)

mattbenjamin:
@cbodley @adamemerson I think this passed
following is a link to all the failures--there are it seems to me now at least 2 different reproducers of the mon start_tick() issue
also a d4n test is sort of taking forever to exit
https://docs.google.com/document/d/1E5CVqEjKyvIKxVEVDDoCDa6Dwt5OLULu208ch3j_TnY/edit?tab=t.0

cbodley (approved):

cbodley:
saw conflicts when cherry-picking https://github.com/ceph/s3-tests/pull/663 to ceph-master, so cherry-picked this commit as well

PR #679, opened 2025-07-29
s3: test list_buckets with MaxBuckets/ContinuationToken

cbodley:
test case for https://github.com/ceph/ceph/pull/64742


cbodley:
cherry-picked to ceph-master

PR #698, opened 2025-09-19
remove boto2 tests under s3tests/ and rename s3tests_boto3

cbodley:
- git rm -rf s3tests
- rm -rf s3tests
- git mv s3tests_boto3 s3tests
- update links to s3tests_boto3 in docs and imports
- remove boto2 dependency from setup.py and requirements.txt


cbodley:
https://pulpito.ceph.com/cbodley-2025-10-14_16:07:01-rgw-wip-cbodley-testing-distro-default-gibba/ shows no new regressions compared to baseline https://pulpito.ceph.com/teuthology-2025-10-10_20:40:24-rgw-main-distro-default-smithi/

cbodley:
> Is there any patch related to this topic in regards to: https://github.com/ceph/ceph/blob/main/src/test/rgw/rgw_multi/conn.py#L1

not yet. similar issue in src/test/rgw/bucket_notification/ tests

we're planning to enable el10 and ubuntu 24 testing soon after the tentacle release, so this will break some test suites

PR #716, opened 2025-12-12
s3tests: add tests for per-bucket cloud transition targeting

mheler:
Add tests to validate the target_by_bucket feature which allows each source bucket to transition objects to a dedicated destination bucket rather than sharing a common target.

New tests:
- test_lifecycle_cloud_transition_target_by_bucket: validates objects land in bucket-specific targets and performs a restore from target
- test_lifecycle_cloud_transition_target_by_bucket_multiple_buckets: validates isolation between different source buckets

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -14,6 +14,7 @@ markers =
     checksum
     cloud_transition
     cloud_restore
+    target_by_bucket
     copy
     encryption
     fails_on_aws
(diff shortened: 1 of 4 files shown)


soumyakoduri:
Could you also please add a test for simple restore when `target_by_bucket` is enabled.

mheler:
> Could you also please add a test for simple restore when `target_by_bucket` is enabled.

I included it as part of the tests now, but I can separate it out if required.

soumyakoduri (approved):

PR #723, opened 2026-01-20
Disable client retries

andy-js:
The boto3 client automatically retries failed S3 operations. For negative test cases—where an error is the expected outcome—this leads to unnecessary delays and can obscure real failures by hiding the initial error. To keep tests fast and deterministic, we should disable automatic retries in the S3 client.


andy-js:
Bump.

PR #728, opened 2026-03-16
s3: Add a test for attempting to complete an unstarted multipart upload

jacquesh:
The [AWS documentation for CompleteMultipartUpload](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html) specifies that the call will return a 404 with the text "NoSuchUpload" to indicate that the specified multipart upload does not exist. This could be the result of an invalid upload ID or a multipart upload that has been aborted or completed.

Of particular significance is the requirement that this API call return a 400-series error (indicating a client-side problem) and not a 500-series error (indicating a server-size problem).
At the time of writing, Ceph returns a 500 in this scenario which may be handled incorrectly by client SDKs and is likely to be misinterpreted by Ceph monitoring as an increase in errors that require operator intervention to fix.

Since this test currently fails, I've marked it as `fails_on_rgw` and created [a tracker](https://tracker.ceph.com/issues/75534) for the issue.

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -5992,6 +5992,7 @@ def test_multipart_upload_empty():
     assert status == 400
     assert error_code == 'MalformedXML'
 
+@pytest.mark.fails_on_dbstore
 def test_multipart_upload_complete_without_create():
     bucket_name = get_new_bucket()
     client = get_client()
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -5992,7 +5992,6 @@ def test_multipart_upload_empty():
     assert status == 400
     assert error_code == 'MalformedXML'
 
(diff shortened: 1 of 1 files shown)


s3tests/functional/test_s3.py
```diff
@@ -5992,6 +5992,25 @@ def test_multipart_upload_empty():
     assert status == 400
     assert error_code == 'MalformedXML'
 
+@pytest.mark.fails_on_rgw # TODO: Remove fails_on_rgw when https://tracker.ceph.com/issues/75534 is fixed
```
cbodley: thanks very much for the test case. would you kindly remove this `fails_on_rgw` marker so i can use this to validate the fix in https://github.com/ceph/ceph/pull/68186?

jacquesh: Gladly! Removed.

PR #736, opened 2026-04-16
sns: test sns:ListTopics with non-root account user

cbodley:
test that identity policy can grant access to ListTopics

Fixes: https://tracker.ceph.com/issues/74595

### DIFF
--- a/s3tests/functional/test_sns.py
+++ b/s3tests/functional/test_sns.py
@@ -3,10 +3,12 @@ import pytest
 from botocore.exceptions import ClientError
 from . import (
     configfile,
+    get_iam_path_prefix,
     get_iam_root_client,
     get_iam_alt_root_client,
     get_new_bucket_name,
     get_prefix,
+    make_iam_name,
     nuke_prefixed_buckets,
 )
 from .iam import iam_root, iam_alt_root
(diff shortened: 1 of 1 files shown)


s3tests/functional/test_sns.py
```diff
@@ -101,6 +103,27 @@ def test_cross_account_topic(sns, sns_alt):
...
+    user = iam_root.create_user(UserName=user_name, Path=path)['User']
+    user_arn = user['Arn']
+    key = iam_root.create_access_key(UserName=user_name)['AccessKey']
+
+    sns = get_iam_root_client(service_name='sns',
+                              aws_access_key_id=key['AccessKeyId'],
+                              aws_secret_access_key=key['SecretAccessKey'])
+
+    # expect AccessDenied for lack of identity policy
+    e = assert_raises(ClientError, sns.list_topics)
```
yuvalif, line 121: shouln't the account root have implicit acces to all resources, including topics?
i think that this is the case for the tenanted user, it can list all of the topics it created.

cbodley: yes, the account root user has this permission by default. but this test case is creating a different (non-root) user and using its credentials to issue this request

yuvalif: can we also verify that the root can list the topics without any policy?

cbodley: the existing `test_cross_account_topic()` test case just above this one is already doing a successful `list_topics()` as an account root user

yuvalif (approved):

PR #740, opened 2026-04-22
rgw-bucket-logging: handle presigned SigV2 requests

nbalacha:
Adds a test to check that the log record AuthenticationType is correctly set to QueryString for SigV2 presigned URLs.

Fixes: https://tracker.ceph.com/issues/76206


yuvalif:
the test does not cover the bucket logging aspects

yuvalif (approved):

yuvalif:
this test is used in order to verify: https://github.com/ceph/ceph/pull/68540

PR #745, opened 2026-05-06
Test return code when getting obj with long version id 

jeanielamwh:
For https://github.com/ceph/ceph/pull/66190#issuecomment-3755420773 which fixes https://tracker.ceph.com/issues/72502.


jeanielamwh:
(Not sure why I can't add inline comment, it says "Error Failed to save comment: An internal error occurred, please try again..")

Anyway, once we have the fix in https://github.com/ceph/ceph/pull/66190, the assertions should be

```python
    assert status == 400
    assert error_code == 'InvalidArgument'
```
Should we merge this 400 version instead? Or do we usually wait until the fix is merged?

PR #747, opened 2026-05-07
s3tests: add test for cloud transition of non-ASCII keys

mheler:
Add a test to validate that lifecycle cloud-tier transition correctly transfers objects whose keys contain non-ASCII bytes to S3-compatible endpoints.

New test:
- test_lifecycle_cloud_transition_non_ascii_key: validates an object with a non-ASCII key transitions and lands at the cloud target

s3 tests for PR: https://github.com/ceph/ceph/pull/68784

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -9857,6 +9857,56 @@ def test_lifecycle_cloud_transition():
         assert status == 404
         assert error_code == 'NoSuchKey'
 
+@pytest.mark.lifecycle
+@pytest.mark.lifecycle_transition
+@pytest.mark.cloud_transition
+@pytest.mark.fails_on_aws
+@pytest.mark.fails_on_dbstore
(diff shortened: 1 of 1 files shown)


soumyakoduri (approved):

s3tests/functional/test_s3.py
```diff
@@ -9857,6 +9857,56 @@ def test_lifecycle_cloud_transition():
...
+    time.sleep(2 * lc_interval)
+    if target_path is None:
+        target_path = "rgwx-default-" + cloud_sc.lower() + "-cloud-bucket"
+    prefix = bucket_name + "/"
+
+    cloud_client = get_cloud_client()
+
+    time.sleep(12 * lc_interval)
+    verify_object(cloud_client, target_path, prefix + non_ascii_key,
+                  non_ascii_key, target_sc)
```
soumyakoduri, line 9908: For sanity check, should we include restore as well?

PR #749, opened 2026-05-13
test: verify Requester field is the assumed-role ARN

ShreeJejurikar:
Add a bucket logging test that asserts the standard access log **Requester** field is set to the assumed-role ARN (per [AWS S3 spec](https://docs.aws.amazon.com/AmazonS3/latest/userguide/LogFormat.html#log-record-fields)) when a request is made with STS temporary credentials.

The test:
1. Creates an IAM role with a trust policy allowing the alt user to call `sts:AssumeRole`, and attaches an inline `s3:*` policy
2. Calls `sts.AssumeRole`, builds a boto3 S3 client from the temporary credentials
3. With the assumed-role client: creates source + log buckets, sets the log-bucket policy, enables Standard bucket logging on the source bucket, then PUTs one object
4. Flushes logs and verifies the resulting log record's Requester field starts with `arn:aws:sts::` and contains `assumed-role/<role>/<session>`

Reuses existing helpers (`get_iam_client`, `get_sts_client`, `_set_log_bucket_policy_tenant`, `_flush_logs`, `_parse_log_record`, `nuke_role`) — no new helper code added.

Pairs with the fix in https://github.com/ceph/ceph/pull/68887. The test will fail in CI against `main` until that lands.

Tracker: https://tracker.ceph.com/issues/71742

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -38,7 +38,7 @@ from .utils import _get_status
 
 from .policy import Policy, Statement, make_json_policy
 
-from .iam import iam_root
+from .iam import iam_root, nuke_role
 
 from . import (
     configfile,
@@ -97,6 +97,9 @@ from . import (
     get_restore_processor_period,
     get_read_through_days,
     create_iam_user_s3client,
+    get_iam_client,
+    get_sts_client,
+    get_parameter_name,
     )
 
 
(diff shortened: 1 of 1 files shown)


s3tests/functional/test_s3.py
```diff
@@ -15816,6 +15819,92 @@ def capture_request_id(**kwargs):
         assert _verify_record_field(body, src_bucket_name, 'REST.PUT.OBJECT', key, "Standard", "RequestID", request_id)
 
 
+@pytest.mark.bucket_logging
+@pytest.mark.fails_on_aws
+@pytest.mark.fails_on_dbstore
+def test_bucket_logging_requester_assumed_role():
+    """verify the standard access log Requester field contains the assumed-role ARN
+    for requests made with STS temporary credentials
+    https://tracker.ceph.com/issues/71742"""
```
yuvalif: probably no need to mention the tracker here

yuvalif (approved):

yuvalif:
test pass. see: https://github.com/ceph/ceph/pull/68887#issuecomment-4536640246

yuvalif:
@ShreeJejurikar thanks for your contribution!
could you please also backport to `ceph-master`?

ShreeJejurikar:
@yuvalif 

> could you please also backport to `ceph-master`?

Done

PR #750, opened 2026-05-13
test/sts: add test for global OIDC provider fallback and precedence

kchheda3:
Add test_assume_role_with_global_oidc_provider_local_override() which validates three scenarios:
1. AssumeRoleWithWebIdentity succeeds via global OIDC provider when no account-level provider exists (fallback path)
2. Account-level provider with wrong client_id causes failure, proving account-level takes precedence over global (client_id validation is always enforced regardless of TLS)
3. After deleting account-level provider, fallback to global resumes

Requires the test task (qa/tasks/s3tests.py) to pre-create a global OIDC provider via 'radosgw-admin oidc-provider create' (done by qa/tasks/s3tests.py in the companion ceph PR #[68850](https://github.com/ceph/ceph/pull/68850)) 

Verified locally with Keycloak 11.0.

### DIFF
--- a/s3tests/functional/test_sts.py
+++ b/s3tests/functional/test_sts.py
@@ -2166,3 +2166,183 @@ def test_get_caller_identity_after_assume_role():
   assert response['Account'] == account_id
   assert response['Arn'] == assume_role['AssumedRoleUser']['Arn']
 
+@pytest.mark.webidentity_test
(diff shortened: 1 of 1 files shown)


kchheda3:
tests moved to actual PR https://github.com/ceph/ceph/pull/68850

PR #729, opened 2026-03-25
test: add test for standard access log request ID

cheese-cakee:
## Summary
- Adds `test_bucket_logging_request_id` to verify that the `RequestID` field in the standard access log matches the `x-amz-request-id` returned in the S3 PUT object response.

## Related
- ceph/ceph#67984
- http://tracker.ceph.com/issues/75416

PR #752, opened 2026-05-15
test_s3: fix test_bucket_logging_request_id for newer botocore

cheese-cakee:
The test `test_bucket_logging_request_id` (introduced in #729) registers
a boto3 `after-call.s3.PutObject` event handler that incorrectly accesses
`kwargs["response"]`.  Botocore does not pass a `response` kwarg in
`after-call` events; the correct kwargs are `parsed`, `http_response`,
`model`, and `context`.

This causes the test to fail in teuthology with:

    KeyError: "response"
    s3tests/functional/test_s3.py:15754: KeyError

Fix by reading the `x-amz-request-id` header directly from the
`put_object()` response, which is simpler and works across all botocore
versions.

Fixes: https://tracker.ceph.com/issues/76587
Related: https://github.com/ceph/s3-tests/pull/729

### DIFF
--- a/s3tests/functional/test_s3.py
+++ b/s3tests/functional/test_s3.py
@@ -15795,14 +15795,9 @@ def test_bucket_logging_request_id():
     assert response['ResponseMetadata']['HTTPStatusCode'] == 200
 
     # capture x-amz-request-id from the put_object response
-    request_id = None
-    def capture_request_id(**kwargs):
-        nonlocal request_id
-        request_id = kwargs['response']['ResponseMetadata']['HTTPHeaders'].get('x-amz-request-id')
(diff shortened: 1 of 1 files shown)


cheese-cakee:
@yuvalif this is the same fix as #751, backported to master. The broken test exists on both branches.

yuvalif (approved):

yuvalif:
teutholgy run: https://pulpito-ng.ceph.com/runs/yuvalif-2026-05-18_09:44:31-rgw:bucket-logging-main-distro-default-trial

yuvalif:
teuthology rerun: https://pulpito-ng.ceph.com/runs/yuvalif-2026-05-18_20:04:00-rgw:bucket-logging-main-distro-default-trial 
pass

PR #759, opened 2026-06-18
rgw: test HEAD bucket object count excludes incomplete multipart parts

BBoozmen:
Add test_head_bucket_usage_multipart_incomplete to verify that X-RGW-Object-Count (returned by HEAD Bucket with ?read-stats=true) does not count in-progress multipart upload parts as completed objects.

After uploading 2 parts without completing, HEAD bucket should report 0 objects and ListObjectsV2 should return 0 keys.

Tests: https://tracker.ceph.com/issues/77509


BBoozmen:
Moved the testcase under ceph repo at https://github.com/ceph/ceph/pull/69591

PR #760, opened 2026-06-24
test_s3: additional checks for versioned buckets

nbalacha:
Verify IsLatest values with delete markers.


nbalacha:
Handled by https://github.com/ceph/ceph/pull/69983

